PR #26, opened 2026-06-23
tinygrad: run an unmodified tinygrad model on the ANE (aneforge.tinygrad)

sbryngelson:
`aneforge.tinygrad.trace(model, input_shape)` runs an unmodified [tinygrad](https://github.com/tinygrad/tinygrad) model on the Neural Engine. tinygrad and ANEForge are both lazy tensor graphs, so the tracer maps the model's ops onto an ANEForge graph instead of computing; that graph compiles to one e5rt program and runs on the engine.

```python
from aneforge.tinygrad import trace
run = trace(model, (1, 3, 224, 224))   # model is plain tinygrad
y = run(x)                             # one ANE dispatch; run.on_ane / run.reason report the path
```

**Sound by construction.** Any op the tracer does not map, or any compile failure, falls the whole model back to tinygrad, so correctness never depends on coverage. What falls back is a hardware limit (the ANE kernel width <= 15, so a 16x16 patch conv), a not-yet-mapped op, or something that cannot be a static graph (host reads, dynamic shapes).

**It runs real models.** tinygrad's own `extra/models`, unmodified, on an M5 Pro vs tinygrad's JIT'd METAL:

| model | on the ANE | cosine | ANE / METAL |
| --- | --- | --- | --- |
| ResNet-18 | full | 1.0000 | 0.25 / 2.03 ms (8.2x) |
| ResNet-34 | full | 1.0000 | 0.38 / 3.83 ms (10.1x) |
| ResNet-50 | full | 0.9998 | 0.46 / 3.85 ms (8.4x) |
| ViT-Ti/16 encoder | full | 1.0000 | 0.61 / 5.60 ms (9.2x) |

Same ANE program ANEForge already benchmarks, so it inherits the energy too (ResNet-18 ~2 mJ on the ANE vs ~35 mJ on the GPU).

tinygrad is an optional dependency (`pip install "aneforge[tinygrad]"`); the module imports without it. `examples/tinygrad_bridge.py` is a runnable demo. The full op coverage and a standalone test suite live in a separate dev repo; here it ships as the `aneforge.tinygrad` adapter alongside the LangChain one. bench JSONs untouched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,11 @@ Then browse [`examples/`](examples/), starting with
 [`examples/quickstart.py`](examples/quickstart.py). For retrieval,
 [`examples/rag_embeddings.py`](examples/rag_embeddings.py) is a LangChain
 `Embeddings` drop-in backed by the on-ANE encoder (4-5x faster than the GPU,
-cosine 1.0000).
+cosine 1.0000). For framework interop, [`aneforge.tinygrad`](aneforge/tinygrad.py)
+runs an unmodified [tinygrad](https://github.com/tinygrad/tinygrad) model on the engine
+(`trace(model, shape)` fuses the forward into one ANE program; tinygrad's ResNet-18/34/50
+and a ViT encoder trace to cosine 1.0000 at 8-10x vs its JIT'd METAL), see
+[`examples/tinygrad_bridge.py`](examples/tinygrad_bridge.py).
 
 ## How it compares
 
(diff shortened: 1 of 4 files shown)


sbryngelson:
Closing. ANEForge and tinygrad are peer frameworks: both build tensor graphs and train models from Python. A tinygrad model is just code against its API, not a portable artifact, so tracing it into an ANEForge graph adds a framework dependency without a real use case. Keeping ANEForge standalone.

PR #119, opened 2026-07-31
einsum: diagonal extraction via mask-and-reduce

### LINKED ISSUE #107: einsum: diagonal extraction via mask-and-reduce
`aneforge/einsum.py` rejects a repeated index within an operand ('ii->i', 'bii->bi') as "needs gather -- unsupported". It does not need gather: multiply by a baked identity mask over the repeated pair and sum one of the axes -- `(x * eye).sum(1)` IS the diagonal, from ops the lowering already has. Implement the rewrite in `_reduce_repeats_check`'s place for the extractable cases (each repeated index appears exactly twice; keep diagonal-WRITE and triple repeats rejected).

Property-test against `np.einsum` over 'ii->i', 'ii->', 'bii->bi', and a mixed spec like 'ii,ij->j'; reject cases keep their current clear errors.

axiom-of-choice:
Fixes #107.

## Summary

`einsum` rejected a doubly-repeated index within one operand (`'ii->i'`, `'bii->bi'`) as "needs gather -- unsupported". It doesn't need gather. Multiplying by a baked identity over the repeated pair and summing one of the axes is the diagonal, from ops the lowering already has.

`_extract_diagonals` does that rewrite on each operand before contraction, so the pairwise contraction and the single-operand transpose/reduce path only ever see operands whose indices are distinct. It loops until no doubly-repeated index remains, so multiple pairs in one operand (`'iijj->ij'`) work, and it returns the same `Tensor` object unchanged when there are no repeats.

The rewrite composes with the rest of the pipeline for free: `'ii,ij->j'` takes the diagonal, then contracts through the existing matmul path.

One consequence worth stating. The identity mask zeroes every off-diagonal term, so each output element keeps exactly one survivor and the reduce adds only zeros. A pure diagonal extraction is bit-exact in fp16, not just within tolerance. The test asserts `array_equal` rather than a relative-error bound.

## High-level changes

`aneforge/einsum.py` (+50 / -12)

- `_diag_pairs(sub)`: maps each index repeated exactly twice to its two positions. Higher multiplicity raises `EinsumUnsupported`.
- `_extract_diagonals(t, sub)`: `(x * eye).sum(q).squeeze(q)` per repeated pair, with the identity reshaped to size-1 on the untouched axes so it broadcasts.
- Wired into `einsum` in place of the per-operand `_reduce_repeats_check` call. That function stays as an internal invariant guard, and its message now says so, so a future path that skips the rewrite fails loudly instead of computing a wrong result.
- Docstrings on `EinsumUnsupported` and `einsum` updated to describe what is actually rejected now.
- `_selftest`: the 3 diagonal cases move from `rejected` into the battery, plus 4 new compositions. `triple-repeat` added to `rejected`.

`tests/test_einsum_diagonal.py` (new, 24 tests): off-device shape and rewrite tests, reject tests, and on-device numerics against `np.einsum`.

## Evaluation

All on-device work run locally on Apple M2 Pro, macOS 26.5.2. CI can't reach an ANE, per CONTRIBUTING.

Self-test (`python aneforge/einsum.py`): 33/33 supported patterns correct on the ANE, 2/2 unsupported correctly rejected.

```
  diag                 ii->i             OK  relerr 0.00000  shape (5,)
  trace                ii->              OK  relerr 0.00000  shape (1,)
  diag-batch           bii->bi           OK  relerr 0.00000  shape (3, 5)
  diag-middle          ibi->bi           OK  relerr 0.00000  shape (3, 4)
  two-diag-pairs       iijj->ij          OK  relerr 0.00000  shape (3, 4)
  diag-then-contract   ii,ij->j          OK  relerr 0.00025  shape (6,)
  diag-both-operands   ii,jj->ij         OK  relerr 0.00032  shape (4, 5)
```

The five pure extractions are exactly 0.00000. The mixed specs carry only the normal fp16 matmul error of the contraction they feed.

| Check | Result |
|---|---|
| `tests/test_einsum_diagonal.py` | 24 passed |
| `test_einsum_ellipsis.py` + new file | 31 passed |
| `test_shapes`, `test_broad`, `test_op_coverage`, `test_corners` | 78 passed |
| `tests/run_corpus.py` | GATE: GREEN, 90/90, 0 red |
| `ruff check` / `.githooks/pre-commit` | clean |

The tests fail without the fix. I verified this by probing `main` through the public API, since a bare `git stash` run only produces a collection error from the new helper imports and proves nothing. All 9 behavioral equations (`ii->i`, `ii->`, `bii->bi`, `bii->b`, `ibi->bi`, `iijj->ij`, `ii,ij->j`, `ii,jj->ij`, `bii,bij->bj`) raise `EinsumUnsupported` on `main` and pass here.

Rejects preserved, as the issue asks:

- `'ij->ii'` (diagonal-WRITE) raises `EinsumUnsupported`, existing message unchanged.
- `'iii->i'` (triple repeat) raises `EinsumUnsupported` with a new explicit message.
- `'ii->i'` on a non-square `(5, 6)` operand raises `ValueError` naming both dims.

## One out-of-scope line

The self-test listed `ellipsis` (`'...ij->...ji'`) under `rejected`, but ellipsis has been supported since `_expand_ellipsis` landed. So `python aneforge/einsum.py` exits nonzero on current `main`, before this change. I confirmed that on a clean tree. It sits in the same list this PR edits and would otherwise keep the self-test red, so I moved it to the battery, where it passes. Happy to split it out if you'd rather keep this PR to #107 alone.

## Risk assessment

Low. The change is additive on a path that previously always raised, so no equation that worked before takes a different route. The only behavior change is that a formerly-rejected class now computes. Contained to `einsum.py`, no ops added, no lowering touched.

- Cost is one `mul` plus one `reduce_sum` per repeated pair, on-engine and inside the same fused program.
- The mask is a baked `const_array` of size `n^2` per pair, folded at compile time.
- `_extract_diagonals` also guards the mismatched-dims case that was previously unreachable, so a malformed `'ii'` over unequal dims gets a clear `ValueError` instead of a shape error deeper in the graph.
- Rank never increases (the mask is rank-equal to the operand and the reduce drops an axis), so the rank<=5 ANE cap isn't approached.

## Pre-merge steps

None. No new dependency, no API break, `coverage` and the dylib untouched.

## Post-merge steps

None required. Two optional follow-ups, left out to keep the diff minimal:

1. `'ij->ii'` (diagonal-write / broadcast-back) is still unsupported and is the natural next step if you want it.
2. `_selftest` duplicates what `tests/` covers and nothing in CI runs it, which is why the stale `ellipsis` entry went unnoticed. Folding it into the pytest suite would stop that recurring. I can open a separate issue.

### DIFF
--- a/aneforge/einsum.py
+++ b/aneforge/einsum.py
@@ -115,7 +115,12 @@ def _diag_pairs(sub: str) -> dict[str, tuple[int, int]]:
     raise EinsumUnsupported(
       f"einsum: index {over!r} appears {counts[over[0]]} times in operand '{sub}'; only a "
       f"doubly-repeated index reduces to a mask-and-reduce diagonal - unsupported on the ANE.")
-  return {ch: tuple(i for i, c in enumerate(sub) if c == ch) for ch, c in counts.items() if c == 2}
+  out: dict[str, tuple[int, int]] = {}
+  for ch, c in counts.items():
+    if c != 2: continue
+    p, q = (i for i, x in enumerate(sub) if x == ch)
+    out[ch] = (p, q)
+  return out
 
 
 def _extract_diagonals(t: Tensor, sub: str) -> tuple[Tensor, str]:
--- a/aneforge/einsum.py
+++ b/aneforge/einsum.py
@@ -8,11 +8,11 @@ from collections import Counter
 
 import numpy as np
 
-from aneforge.graph import Tensor
+from aneforge.graph import Tensor, _const
 
 
 class EinsumUnsupported(NotImplementedError):
-  """Raised for equations not lowerable to ANE matmul/reduce ops (diagonal/trace, ellipsis, invented output dims)."""
+  """Raised for equations not lowerable to ANE matmul/reduce ops (diagonal-write, triple repeats, ellipsis)."""
 
 
 # equation parsing
@@ -99,13 +99,42 @@ def _parse(equation: str, n_operands: int) -> tuple[list[str], str]:
 
 
 def _reduce_repeats_check(sub: str) -> None:
-  """A repeated index within one operand is a diagonal extraction (gather)."""
+  """Invariant guard: operands reach the contraction with distinct indices (`_extract_diagonals` ran first)."""
   if len(set(sub)) != len(sub):
     dup = sorted({ch for ch in sub if sub.count(ch) > 1})
     raise EinsumUnsupported(
-      f"einsum: repeated index {dup} within operand '{sub}' is a diagonal/trace "
-      f"extraction, which needs gather - unsupported on the ANE. "
-      f"(Reject rather than return wrong results.)")
+      f"einsum: repeated index {dup} within operand '{sub}' reached the contraction "
+      f"undiagonalized (internal invariant).")
+
+
+def _diag_pairs(sub: str) -> dict[str, tuple[int, int]]:
+  """Map each index repeated exactly twice in `sub` to its two positions; a higher multiplicity rejects."""
+  counts = Counter(sub)
+  over = sorted(ch for ch, c in counts.items() if c > 2)
+  if over:
+    raise EinsumUnsupported(
+      f"einsum: index {over!r} appears {counts[over[0]]} times in operand '{sub}'; only a "
+      f"doubly-repeated index reduces to a mask-and-reduce diagonal - unsupported on the ANE.")
+  return {ch: tuple(i for i, c in enumerate(sub) if c == ch) for ch, c in counts.items() if c == 2}
+
+
+def _extract_diagonals(t: Tensor, sub: str) -> tuple[Tensor, str]:
+  """Rewrite each doubly-repeated index in one operand as its diagonal, dropping the second axis.
+
+  No gather needed: `(x * eye).sum(axis)` IS the diagonal, from ops the lowering already has. The baked
+  identity zeroes every off-diagonal term, so each output element keeps exactly one survivor - the
+  reduce adds only zeros and the extraction is bit-exact in fp16 (unlike a true multi-term sum)."""
+  while True:
+    pairs = _diag_pairs(sub)
+    if not pairs: return t, sub
+    ch, (p, q) = min(pairs.items())
+    n = t.shape[p]
+    if t.shape[q] != n:
+      raise ValueError(f"einsum: repeated index {ch!r} in operand '{sub}' spans dims "
+               f"{n} and {t.shape[q]}; a diagonal needs them equal")
+    mshape = [n if i in (p, q) else 1 for i in range(len(sub))]   # eye over the pair, size-1 elsewhere
+    t = (t * _const(np.eye(n, dtype=np.float16).reshape(mshape))).sum(q).squeeze(q)
+    sub = sub[:q] + sub[q + 1:]
 
 
 # operand-level helpers
@@ -214,7 +243,7 @@ def _expand_to(t: Tensor, present: list, target: list, _sizes) -> Tensor:
 
 def einsum(equation: str, *operands: Tensor) -> Tensor:
   """numpy-style einsum lowered to aneforge ops (matmul-reducible patterns, incl. '...' batch
-  dims); rejects diagonal/trace and repeated output indices."""
+  dims and diagonal/trace extraction); rejects diagonal-WRITE and repeated output indices."""
   if not operands:
     raise ValueError("einsum: at least one operand is required")
   if not all(isinstance(o, Tensor) for o in operands):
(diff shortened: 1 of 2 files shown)


axiom-of-choice:
Flagging one thing in this diff that isn't part of #107, so you don't have to discover it in review.

`_selftest` listed `ellipsis` (`'...ij->...ji'`) under `rejected`, but ellipsis has been supported since `_expand_ellipsis` landed. That entry is stale, and the effect is that `python aneforge/einsum.py` already exits nonzero on a clean `main`, independent of this PR:

```
$ git stash && python -c "import aneforge as af; from aneforge.einsum import einsum; print(einsum('...ij->...ji', af.input((2,3,4))).shape)"
(2, 4, 3)          # computes fine, so the reject entry is what's wrong
```

It sits in the same `rejected` list this PR edits, so leaving it would have kept the self-test red at 2/3 while my new diagonal cases passed. I moved it into the battery, where it passes on device. That's 2 lines of the diff.

Happy to pull it into its own commit or a separate PR if you'd rather keep this one strictly to #107. Say which and I'll re-push.

Related, and the reason it rotted unnoticed: `_selftest` duplicates coverage that `tests/` owns, and nothing in CI runs it. That's reasonable given most of it needs an ANE, though the reject half is pure host-side and would run off-device fine. I can open a separate issue to fold that half into the pytest suite so a stale entry fails visibly. Not doing it here to keep the diff minimal.

aneforge/einsum.py
```diff
@@ -99,13 +99,42 @@ def _parse(equation: str, n_operands: int) -> tuple[list[str], str]:
...
+
+def _diag_pairs(sub: str) -> dict[str, tuple[int, int]]:
+  """Map each index repeated exactly twice in `sub` to its two positions; a higher multiplicity rejects."""
+  counts = Counter(sub)
+  over = sorted(ch for ch, c in counts.items() if c > 2)
+  if over:
+    raise EinsumUnsupported(
+      f"einsum: index {over!r} appears {counts[over[0]]} times in operand '{sub}'; only a "
+      f"doubly-repeated index reduces to a mask-and-reduce diagonal - unsupported on the ANE.")
+  return {ch: tuple(i for i, c in enumerate(sub) if c == ch) for ch, c in counts.items() if c == 2}
```
sbryngelson: The `types (pyright)` check fails here. `tuple(i for i, c in enumerate(sub) if c == ch)` is inferred as variable-length `tuple[int, ...]`, which pyright won't narrow to the declared `-> dict[str, tuple[int, int]]` return type:

```
einsum.py:118 - error: Type "dict[str, tuple[int, ...]]" is not assignable to return type "dict[str, tuple[int, int]]" (reportReturnType)
```

The runtime is correct (a count-2 index always yields exactly two positions) — pyright just can't prove the length. Unpacking into a pair first fixes it while keeping the precise `tuple[int, int]` type that `_extract_diagonals`' `ch, (p, q) = ...` relies on. Verified locally: pyright clean and all 24 tests in `test_einsum_diagonal.py` still pass.

```suggestion
  pairs: dict[str, tuple[int, int]] = {}
  for ch, c in counts.items():
    if c == 2:
      p, q = (i for i, ci in enumerate(sub) if ci == ch)   # exactly two positions
      pairs[ch] = (p, q)
  return pairs
```

sbryngelson (commented):
Nice PR — the mask-and-reduce diagonal (no gather) is a clean fit for the engine, and the bit-exactness reasoning holds up: I ran the suite on-ANE and the pure-diagonal cases are exactly bit-identical to numpy (relerr 0.0), with the full 33-pattern battery correct and both reject cases firing. One inline note: the only failing CI check is `types (pyright)` on a return-type annotation — left a suggestion with a verified one-block fix. Everything else is green.

aneforge/einsum.py
axiom-of-choice: Applied, thanks for the verified fix. Your diagnosis is exactly right: the count-2 filter guarantees two positions at runtime, pyright just can't prove the length through tuple(genexp).

I'd independently pushed the same unpacking in https://github.com/sbryngelson/ANEForge/commit/9292da3eafdbf45bbd645c159b6857379a637cee before your note landed, so https://github.com/sbryngelson/ANEForge/commit/bc77d461c3d644e635d0036e38477a54ce85aefd switches to your wording verbatim (pairs, the if c == 2 nesting, and the ci loop var) rather than leaving my near-identical variant in place. Your repo, your naming.

Re-verified on M2 Pro / macOS 26.5.2 after the change:

```
pyright aneforge/einsum.py: 0 errors
tests/test_einsum_diagonal.py: 24 passed
self-test: 33/33 on-device, 2/2 rejects
ruff + .githooks/pre-commit: clean
```
Also good to hear the bit-exactness reproduced on your chip, since that was the part I most wanted a second machine on. Thanks for running the battery.

Issue #137, opened 2026-08-01
Help map the ANE roofline across every Apple Silicon chip (2-minute run, you get credited)


sbryngelson:
We are characterizing the Apple Neural Engine's rooflines across the whole Apple Silicon line -- and we need runs from machines we do not have. **If you have any Mac with an M-series chip, a short command adds your machine to the map, and you get credited by GitHub handle.**

## Why this matters

The ANE behaves differently across silicon generations and even across chassis of the *same* chip (a fanless MacBook Air throttles where a Mac mini does not). Two things we are mapping:

- **Performance rooflines** -- peak fp16 compute, memory bandwidth, ridge point, perf/watt, decode throughput.
- **Numeric correctness cliffs** -- the magnitudes at which fp16 math on the engine silently returns the wrong answer. These are *silicon-dependent*, and the collected data already shows it: the slice crop-DMA clamp at `|value| > 4094` is present on M1 / M1 Max / M2 Pro but **gone by M5** -- the pre-A16 -> A16 transition, now bracketed by real submissions. (Background: #115, `docs/cross-chip.md`.)

## How to submit

The tooling is live on `main`. From a clean checkout, pick your level:

```bash
# 1. cliffs only -- fastest, no sudo, works on battery (~2 min):
PYTHONPATH=. python3 bench/roofline_suite.py

# 2. RECOMMENDED -- cliffs + headline perf (peak GEMM, bandwidth, ridge, perf/W, decode).
#    A few minutes; prompts once for sudo to read watts (required for --perf). Battery + High Power is fine:
PYTHONPATH=. python3 bench/roofline_suite.py --perf

# then regenerate the table and open a PR with both files:
python3 bench/aggregate_rooflines.py
git add bench/results/rooflines/roofline-*.json bench/results/ROOFLINES.md
```

Open the PR and you are on the board. (`--perf-full` runs the complete paper-grade battery, ~30 min -- only needed for paper-quality reproduction, not for contributing a datapoint.)

[...]

Issue #115, opened 2026-07-23
matmul saturates at fp16_max/2 (~32752); two transpose-fed cases return inf even below that


sbryngelson:
Three matmul/reduce accumulation behaviors characterized by the differential fuzzer (scripts/fuzz.py) and controlled sweeps on M5 / macOS 26.5, at opt=0.

**1. Matmul output saturates at ~32752 = fp16_max/2, for every contraction size K.** All-1000 inputs with ones weights: K=32 (sum 32000) exact, K=33 (sum 33000) returns inf. Same threshold at K=4/8/64: 32400 exact, 33280/33600/35200 all inf. The result format evidently carries a spare factor of two - the matmul sibling of the documented slice-x16 saturation at 4094. Practical rule: any matmul whose true output magnitude exceeds ~32K silently returns inf, even though fp16 represents values to 65504.

**2. Integer reduce sums lose bit-exactness crossing 2048.** A reshape->reduce_sum graph of small integers whose row sum is exactly 2048 returned 2050 (fuzzer finding e5afb32f) - consistent with fp16 rounding of partials at the 2048 spacing boundary somewhere in the reduce path, while sums below 2048 are bit-exact.

**3. OPEN: two transpose-fed matmuls return inf with mathematically bounded sums.** Fuzzer findings 818c07b3 ([31,15]@[15,16]) and f301582d (slice->reshape->transpose->[6,31]@[31,16]) return inf although max_j sum_k |x_jk*w_kj| = 17321 - no summation order can exceed the saturation threshold. Both replayable via `python3 scripts/fuzz.py --repro finding-<fp>.json` (specs attached in the fuzz report; both carry 30000-magnitude danger values through layout ops into the matmul). This one is not yet explained by the saturation rule and needs layout-level investigation.

Items 1-2 are now modeled by the fuzzer's accumulation screen (bounds sit below both cliffs); item 3 escapes correct screens and is the active lead.

sbryngelson:
[...]

PR #145, opened 2026-08-01
bench: roofline datapoint for Apple M2 Pro (Mac mini, cliffs + perf)

axiom-of-choice:
Datapoint for #137. Adds Apple M2 Pro to the map.

Run from a clean checkout of `263e02f` (`dirty: false` in the report), Mac14,12 Mac mini, macOS 26.5.2 (25F84), AC, automatic energy mode.

```
matmul inf-cliff ~32758.8  |  slice cliff 4094.1  |  reduce exact <= 2048
```

## The interesting one: the slice clamp is present on M2 Pro

| Chip | slice-x16 (Q.4 crop-DMA) |
|---|---|
| Apple M2 Pro (this run) | **clamp @ 4094** |
| Apple M5 Pro | exact (no clamp) |

The issue says the pre-A16 crop-DMA clamp is gone by A16/M5, and this is the first M-series confirmation in the table that it *is* there on an earlier chip. So the boundary is now bracketed from both sides rather than only asserted on one.

## The other two look architecture-wide, not per-generation

Both match the M5 Pro:

- **matmul inf-cliff ~32759**, and flat across `K in {4, 8, 32, 64, 128}`, consistent with `fp16_max/2` independent of reduction depth, which is what I measured by hand in #115 before this tooling existed.
- **reduce exactness `<= 2048`**, first wrong at 2049 returning 2050.

Worth noting the second one against history: when I probed the integer reduce by hand on this same machine (#115, 2026-07-31) it looked exact past 2048 and I reported it as *not* reproducing. The suite disagrees, and the suite is right. My hand probe kept the addends small enough that the sum stayed on the exact fp16 grid, so it never actually crossed the boundary the metric measures. The harness picks the boundary case properly. Retracting that half of my earlier report; the per-generation split there was my measurement artifact, not silicon.

## Perf half (added)

Ran `--perf` with sudo from a clean tree. Measured ANE peaks:

| Peak fp16 GEMM | Peak perf/W | Streaming bandwidth |
|---|---|---|
| **3.26 TF/s** | **1379 GFLOP/s/W** | **1.0 GB/s** |

Two table cells stay empty, both for reasons worth stating rather than hiding.

**Ridge, and the table's GEMM/bandwidth cells, are blank because of a harness bug I hit and filed as #146.** `aggregate_rooflines` reads those three from `roofline_analysis.py`, and that step raised `KeyError: 'ane'`. Cause: `layer_norm` failed to compile on the ANE at these shapes, so the bandwidth script correctly recorded `ane_gbps: None` and that archetype's `peak` block has no `ane` key, but `load_ceilings` iterates a hardcoded `("CPU","GPU","ANE")` per archetype and raises. Not specific to this machine: any run where one archetype loses one engine loses the whole synthesis. The measured numbers above are in this report's `perf_rooflines` regardless, so nothing is lost, only unsynthesized. Once #146 lands, re-running `aggregate_rooflines.py` fills those cells with no new measurement.

**Decode is blank for a genuinely different reason, and it is a datapoint.** On M2 the decode model's `[1, 32000]` vocab projection exceeds **ANE family 3's 16384 max dimension**, so both batch rows return:

```
NotImplementedError: aneforge: tensor (1, 32000) (op 'matmul') exceeds ANE family 3's
max dimension 16384; tile the graph or target a higher chip.
```

That is a real per-generation capability limit, not a harness failure. The same model measures fine on your M5 (117 tok/s in the table), so `bench/decode_measurement.py` currently cannot produce a decode number on family-3 silicon at all without tiling the projection. Worth deciding whether the harness should tile it, or whether the table should distinguish "not measured" from "cannot run on this family".

## Not included

Nothing further. Both halves are in.

## Files

- `bench/results/rooflines/roofline-apple-m2-pro-Mac14_12-723396ad9091-1c1aa174.json` (new)
- `bench/results/ROOFLINES.md` (regenerated with `bench/aggregate_rooflines.py`, now 2 machines)


bench/results/ROOFLINES.md
```diff
@@ -1,16 +1,17 @@
 <!-- generated by bench/aggregate_rooflines.py; do not edit by hand -->
 # ANE roofline results
 
-Accumulated across **1 machine(s)**, 1 submission(s). Each row is a distinct machine (grouped by hardware hash; identical silicon in different chassis stays separate by model identifier).
+Accumulated across **2 machine(s)**, 2 submission(s). Each row is a distinct machine (grouped by hardware hash; identical silicon in different chassis stays separate by model identifier).
 
 Add yours: `PYTHONPATH=. python3 bench/roofline_suite.py --contributor <your-gh-handle>`, commit the JSON it writes under `bench/results/rooflines/`, regenerate this file with `python3 bench/aggregate_rooflines.py`, and open a PR. See [`rooflines/README.md`](rooflines/README.md).
 
-Thanks to [@sbryngelson](https://github.com/sbryngelson) for contributing datapoints.
+Thanks to [@axiom-of-choice](https://github.com/axiom-of-choice), [@sbryngelson](https://github.com/sbryngelson) for contributing datapoints.
```
axiom-of-choice, line 8: @sbryngelson Feel free to edit this contribution sections

Issue #155, opened 2026-08-02
compile: let graphs declare input activation ranges so opt=1 can keep int8 when provably safe


sbryngelson:
Split out of #153 as the follow-up that turns its layer 1 into a real optimization rather than a blanket decline.

## Problem

#153's fix declines the int8-weight variant at `opt=1` when the activation magnitude can exceed the variant's Q.4 x16 saturation ceiling (`4094 = fp16_max/16`; see #153/#149). The check is fail-closed on an unknown bound, using the existing precedent at `_targets.py:215`:

```python
if max_abs is None or float(max_abs) > FP16_SLICE_SAT: return "saturation"
```

But `_node_max_abs` (`_compile.py:1096`) bounds **baked constants only** -- it returns `None` for any runtime input:

```python
def _node_max_abs(t: Tensor):
    """Static bound on a node's value magnitude: a baked const's max, else None."""
    v = t.attrs.get("value")
    ...
```

So on any graph with runtime inputs (i.e. almost all of them) the bound is `None`, int8 is declined at `opt=1` **essentially always**, and layer 1 collapses to option 3 ("no lossy variants at opt=1") in practice. That is safe but throws away int8 speedups even when the activations are known to be well under 4094.

## What's missing

A way to bound runtime-input activations statically, in two pieces:

1. **Declare an input range.** Give `af.input(shape, dtype)` an optional activation bound, e.g. `af.input(shape, max_abs=6.0)` (or a `(lo, hi)`), stored on the input node so `_node_max_abs` returns it instead of `None`. Undeclared inputs stay `None` -> still fail-closed.
2. **Propagate the bound** conservatively to each lossy-variant node: elementwise `max|f(x)|`, matmul `|x| @ |W|` upper bound, norms clamping to ~O(1), etc. The `opt=1` ceiling check then compares the *propagated* bound against `Q4_X16_SAT` and **keeps int8 when it is provably below**.

Over-estimation is always safe here -- a loose bound just declines int8 more often, never wrongly keeps it -- so the propagation can start simple (cover the common ops feeding matmuls) and tighten over time.

## Payoff

[...]

PR #156, opened 2026-08-02
opt=1: decline the int8 variant above its activation-encoding ceiling

### LINKED ISSUE #153: autotuner emits inf instead of declining the int8 variant when activations exceed its saturation ceiling
Filing this as its own issue per your request in https://github.com/sbryngelson/ANEForge/issues/115#issuecomment-5153421171 (item 3 follow-up). Splitting the contract question out of the #115 characterization.

## Summary

At `opt=1` the autotuner can pick the int8-weight variant for a matmul whose activations carry magnitudes above that variant's saturation ceiling. The result is `inf` in the output rather than a gated accuracy miss, on a matmul whose true sums are well inside fp16 range. `opt=0` on the same graph is exactly finite.

The right contract, as you put it: **decline the lossy variant when `|activation|` exceeds its saturation ceiling, instead of emitting `inf`.**

## Reproduction

Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2, `d6de531`. `[31,15] @ [15,16]`, standard-normal input with ~1/16 of entries set to 30000, weights `N/sqrt(K)`, chosen so the true product stays bounded:

```
bound |x|@|W| max = 30133.0        (well under the ~32760 fp16_max/2 cliff)
  opt=0: 0/496 non-finite,   max 29280.0
  opt=1: 227/496 non-finite, max 3.3
```

Nearly half the output is `inf`, and the surviving finite values top out at 3.3, so the result is not merely imprecise, it is unusable. From the #115 sweep this is absolute rather than probabilistic: every bounded case with a ~30000 carrier failed at `opt=1` (93/93 direct, 228/228 transpose-fed) and none failed at `opt=0`.

## Why the existing accuracy gate does not catch it

`_optimize.py` does validate variants against the baseline, and an `inf` output would fail that check: with `ref` finite, `relerr = |cur - ref|.max() / denom` is `inf`, which is `> _ACCURACY_TOL`, so the variant is rejected.

The gap is the no-reference path. In `_measure_with_ref` (`_optimize.py:618-630`):

```python
if ref is not None and cur.shape == ref.shape:
    relerr = float(np.abs(cur - ref).max() / denom)
else:
    relerr = float("nan")
```

and `nan > _ACCURACY_TOL` is `False` in IEEE comparison, so a variant with no usable reference passes the accuracy filter by default rather than being rejected. Any later `relerr > tol` style test inherits the same weakness: `nan` never trips a `>` comparison.

## Suggested fix

Two layers, either useful alone:

1. **Refuse the variant up front.** Before accepting an int8 candidate, compare the observed activation magnitude against that variant's saturation ceiling and drop it when exceeded. This is the contract you described and it needs no measurement.
2. **Make the gate fail-closed.** Treat a non-finite or unavailable `relerr` as a rejection rather than a pass, e.g. `if not (relerr <= tol): reject`, which rejects `nan` and `inf` alike. Guards against future lossy variants with the same shape of bug.

I have not sent a patch since the ceiling value per variant is a design decision on your side. Happy to implement either layer if you say which you prefer.

## Context

Split from #115 item 3, which is now understood: the failure is the `opt=1` lossy path, not the layout, and it reproduces on a second generation. Related: #149 (`layer_norm` compile failure), also split out of that investigation.

axiom-of-choice:
Fixes #153. Implements option 1 with the fail-closed rule on an unknown bound, per your design call, plus the three asks that came with it.

## The bug

`opt=1` selects on predicted cost and never measures (`_compile._compile_opt`), so nothing validated the lossy int8 variant. A matmul whose true sums sit inside fp16 range came back with `inf`:

```
[31,15] @ [15,16], true |x|@|W| max = 30133   (fp16 cliff is ~32760)
  opt=0: 0/496 non-finite
  opt=1: 179/496 non-finite       <- int8 variant won on cost, unvalidated
```

## The three asks

**Share the constant, don't add a second 4094.** `_targets.FP16_SLICE_SAT` becomes `Q4_X16_SAT`, naming the encoding rather than the slice op, with the old name kept as an alias so nothing outside breaks. Your invariance result is what justifies this: the ceiling not moving with weight scale, K, N or seed means the int8-weight matmul is routing activations through the same Q.4 x16 encoding, so one constant is correct rather than a coincidence.

**Per-variant, not global.** `_variant_act_ceiling(cfg)` is keyed by config flag and returns `None` for variants with no encoding limit, so int4-LUT and sparse can declare their own ceilings without inheriting int8's. A test pins that `{"int8": False}` and const-fold report `None`.

**Pin 4094 in a regression test.** Done, including the exactness argument: `np.float16(4095) == np.float16(4096)`, so no fp16 activation lies between 4094 and 4096 and `|act| <= 4094` is a gate with no boundary gap. Parametrized either side of it.

## How the bound is computed

`_node_max_abs` only ever bounded a single node, so it could not answer "what is the largest activation in this graph". Added `_act_max_abs`, which walks to the leaves and returns `None` if any leaf is a runtime `input`. `_exceeds_act_ceiling` then treats `None` as unsafe, mirroring the precedent you pointed at (`_targets.py:215`):

```python
if max_abs is None or float(max_abs) > FP16_SLICE_SAT: return "saturation"
```

`_variants(out, drop_unsafe=True)` filters on that, and only the `opt=1` path passes the flag. `opt=2` is untouched, since it validates by measurement and was already safe.

## Verification

Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2.

The repro is fixed, and `opt=1` now matches `opt=0` exactly:

```
  opt=0: 0/496 non-finite, max 29280.0
  opt=1: 0/496 non-finite, max 29280.0      (was 179/496)
```

**It is a gate, not a ban.** int8 still survives at `opt=1` where it is safe:

| graph | static bound | int8 offered |
|---|---|---|
| runtime `input` | `None` | no (fail closed) |
| const-fed, 10.0 | 10.0 | **yes** |
| const-fed, 5000.0 | 5000.0 | no |
| const-fed, 4094.0 | 4094.0 | **yes** |
| const-fed, 4096.0 | 4096.0 | no |

| Check | Result |
|---|---|
| `tests/test_opt1_act_ceiling.py` | 10 passed, all build-level so they run in CI without an ANE |
| `test_routes`, `test_tune_guards`, `test_compile_targets` | 36 passed |
| `ruff` / `pyright aneforge` / pre-commit | clean (pyright's 9 findings are the pre-existing optional-import ones) |

Behaviour gap confirmed on `main` through the public API rather than by a stashed-file run, which only produces a collection error: `_variants` has no `drop_unsafe` parameter there and offers int8 unconditionally.

## On the cost, honestly

You flagged that on any graph with runtime inputs the bound is `None`, so this declines int8 at `opt=1` nearly always, landing close to option 3 in practice. That matches what I see, and I have not measured the throughput given up on real `opt=1` workloads either, so I would not claim the cost is negligible. Two things make me still prefer this shape: it keeps int8 available for const-fed subgraphs today, and it is the right structure once input ranges can be declared, which is the change that makes this meaningfully different from option 3. Happy to switch to option 2 (single probe dispatch, reject on non-finite) if you would rather trade a dispatch for the throughput.

Declaring an input range looks worth its own issue, as you said. Say the word and I will file it with this PR as context.

## Not in this PR

The fail-closed `not (relerr <= tol)` change on the `opt=2` path, which you asked for as a separate small PR since it is independently merge-worthy rather than a fix for this. Sending that next.

### DIFF
--- a/aneforge/_compile.py
+++ b/aneforge/_compile.py
@@ -1093,6 +1093,25 @@ class CrossChipFP16Warning(UserWarning):
   """Emitted by `cross_compile_check` when a graph compiles for a different-family target but an op's fp16 value can diverge."""
 
 
+def _act_max_abs(out: Tensor):
+  """Static bound on the largest activation magnitude feeding `out`, or None if any leaf is unbounded.
+
+  Only baked constants carry a bound; a runtime `input` does not, so a graph with one returns None and
+  callers must treat that as unsafe (see `_optimize._exceeds_act_ceiling`)."""
+  seen, stack, worst = set(), [out], 0.0
+  while stack:
+    t = stack.pop()
+    if id(t) in seen: continue
+    seen.add(id(t))
+    if not t.srcs:                                   # leaf: const (bounded) or input (unbounded)
+      b = _node_max_abs(t)
+      if b is None: return None
+      worst = max(worst, b)
+      continue
+    stack.extend(t.srcs)
+  return worst
+
+
 def _node_max_abs(t: Tensor):
   """Static bound on a node's value magnitude: a baked const's max, else None."""
   v = t.attrs.get("value")
@@ -1247,8 +1266,9 @@ def _compile_opt(out: Tensor, int8: bool, opt):
   """opt=1/2/'max' dispatch (lazy-imported so aneforge imports without the optimizer)."""
   from . import _optimize
   if opt in (1,):
-    # cost-model pick: cheaper-predicted variant, no measurement.
-    cfgs = _optimize._variants(out)
+    # cost-model pick: cheaper-predicted variant, no measurement. Since nothing here validates the
+    # result, drop lossy variants whose activation-encoding ceiling this graph can exceed (#153).
+    cfgs = _optimize._variants(out, drop_unsafe=True)
     best = min(cfgs, key=lambda c: _optimize._estimate_variant(out, c))
     return _optimize.build_variant(out, best)
   if opt in (2, "max", "2"): return _optimize.tune(out)
--- a/aneforge/_optimize.py
+++ b/aneforge/_optimize.py
@@ -273,8 +273,34 @@ def _route_configs(routes) -> list:
   return configs
 
 
-def _variants(out):
-  """Legal variant configs: route flips (lossless) + global int8 / const-fold / scalar-fold (lossy)."""
+# Activation-encoding ceiling per lossy variant: above it the variant saturates to inf rather than
+# losing accuracy. Keyed by config flag so int4-LUT / sparse can declare their own encodings; only
+# the int8-weight matmul is characterized so far (Q.4 x16, see #153).
+def _variant_act_ceiling(cfg) -> "float | None":
+  """The activation magnitude this variant can encode, or None if it has no such limit."""
+  from ._targets import Q4_X16_SAT
+  return Q4_X16_SAT if cfg.get("int8") else None
+
+
+def _exceeds_act_ceiling(out, cfg) -> bool:
+  """True if `cfg` cannot safely encode this graph's activations.
+
+  Fail closed on an unknown bound, matching `_targets.py`'s slice-saturation rule: a runtime input
+  has no static bound, so it must be treated as unsafe rather than assumed small. Without this an
+  unbounded activation silently saturates to inf at opt=1, which never runs the accuracy gate."""
+  ceiling = _variant_act_ceiling(cfg)
+  if ceiling is None: return False
+  from ._compile import _act_max_abs
+  bound = _act_max_abs(out)
+  return bound is None or float(bound) > ceiling
+
+
+def _variants(out, drop_unsafe: bool = False):
+  """Legal variant configs: route flips (lossless) + global int8 / const-fold / scalar-fold (lossy).
+
+  `drop_unsafe` removes lossy variants whose activation-encoding ceiling this graph can exceed. The
+  cost-model-only path (opt=1) needs it because nothing downstream validates the variant; opt=2
+  leaves it off and rejects on measured error instead."""
   configs = _route_configs(_route_ids(out))
   if _has_weights(out):
     configs.append({"int8": True, "decomp": [], "lossy": True})   # PRIMARY: global int8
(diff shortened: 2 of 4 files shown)


sbryngelson:
This is an excellent implementation -- exactly the design we converged on: per-variant `Q4_X16_SAT` (renamed from `FP16_SLICE_SAT` with the alias kept), fail-closed on an unknown bound, int8 retained for bounded graphs, opt=2 untouched, and the 4094 pin + exact-`fp16(4095)==4096` boundary test are precisely what I asked for. Thank you.

I'm going to **hold it**, though -- and it's purely a sequencing call, nothing about the code. As `test_opt1_declines_int8_for_runtime_input` makes explicit, `_act_max_abs` returns `None` for any graph with a runtime `af.input`, so int8 is dropped at opt=1 for **every real inference graph**, not just the large-activation ones that actually overflow. That trades a *narrow* correctness bug (int8 -> inf only above 4094, a rare pathological case) for a *broad* opt=1 throughput regression (int8 off for all runtime-input models), and the perf-recovery path -- #155, declaring input ranges so a bounded graph keeps int8 -- doesn't exist yet.

So let's land **#155 first, then this, together**: provably-bounded and range-declared graphs keep the fast path, and only genuinely-unbounded activations lose int8 at opt=1. The fix is right; I just don't want to ship the regression window in between. #155 is the more interesting piece anyway, and I'll review it as soon as you pick it up.

One nit for when this comes back: the test docstring references `bench/opt1_int8_saturation_repro.py`, which isn't in the PR -- either add it (a committed on-device repro would be great to have) or drop the line.

axiom-of-choice:
Understood on the sequencing, and the reasoning is right: trading a narrow correctness bug for a broad throughput regression is the wrong order when the recovery path does not exist yet.

Picked up #155: **#161**. It carries this gate plus the bound machinery, so it supersedes this PR rather than stacking, and there is no window where int8 is off for everything.

The part that turned out to matter more than the declaration itself: the saturating activations bound an *undeclared* input, so `softmax`, `sigmoid`, `tanh`, `l2_norm` and `relu6` recover int8 at opt=1 with no annotation. Anything downstream of one keeps the fast path for free, which covers a lot of real graphs without users touching their code.

Also fixed your nit there: `bench/opt1_int8_saturation_repro.py` is committed in #161, and the docstring reference is accurate now.

Happy to close this one in favour of #161, or leave it open if you would rather diff the two. Your call.

sbryngelson:
Closing in favor of #161, which carries this same opt=1 ceiling gate plus the #155 bound machinery -- so there's no window where int8 is off for everything, and any graph downstream of a saturating op (softmax/sigmoid/tanh/l2_norm/relu6) keeps the fast path with no annotation. The design here was exactly right; #161 is the same idea with the recovery path attached. Thanks -- see the review over there (one rebase needed so it doesn't revert #157).

Issue #141, opened 2026-08-01
bench: device_compare_wattcomplete aborts on missing examples/vit_demo (ViT-B/16 workload)


sbryngelson:
`bench/device_compare_wattcomplete.py` crashes partway through in a clean checkout. Its `wl_vit_b16()` workload does `sys.path.insert(0, str(REPO / "examples")); import vit_demo as vd` (around line 580), but `examples/vit_demo.py` is not in the repo (its siblings -- `demo.py`, `fft.py`, `gpt_generate_ane.py`, etc. -- are). So the whole script aborts with:

```
ModuleNotFoundError: No module named 'vit_demo'
```

This surfaced running the roofline suite's `--perf-full` (#140): 5 of 6 perf scripts complete, but wattcomplete dies here, so it produces no per-rail-watt result. (The default `--perf` fast path does not run wattcomplete, so contributors are unaffected -- this only bites the full paper-grade battery and manual `python3 bench/device_compare_wattcomplete.py` runs.)

**Fix (good first issue): skip the ViT workload gracefully when its module is absent**, exactly like the sibling workloads already do for a missing optional dep. `wl_vit_b16()` already has the pattern one line up:

```python
    if not HAVE_TV:
        note(wl, "skipped - torchvision unavailable.")
        return
    sys.path.insert(0, str(REPO / "examples"))
    import vit_demo as vd          # <- this raises when examples/vit_demo.py is absent
```

Wrap the import (and load) in a `try/except ImportError` that calls `note(wl, "skipped - examples/vit_demo.py unavailable.")` and returns, mirroring the `not HAVE_TV` / `not HAVE_HF` skips already in this file (search for `note(wl, "skipped`). The rest of the battery then completes and just omits the ViT row.

Reject boundary: do NOT stub fake ViT numbers to fill the row -- skip it cleanly, the way the other unavailable workloads are skipped. Fully *restoring* the ViT-B/16 workload (adding `examples/vit_demo.py` with `build_vit` / `load_vit_weights` / `torch_ref` and the `IMG`/`DIM`/`SEQ`/`N_LAYERS` constants the code expects) is a larger, separate follow-up; this issue is just the graceful skip so the script stops aborting.

Verifying needs any Apple Silicon Mac with the bench extras (`pip install -e ".[bench]"`) to run `python3 bench/device_compare_wattcomplete.py` and confirm it now completes with the ViT row skipped; the code change itself is a small guarded import.

8nt0n:
[...]

PR #160, opened 2026-08-02
Improve import handling for vit_demo

8nt0n:
### Testing
- [x] Verified Python syntax via `python3 -m py_compile`.
- [x] Verified `try/except ImportError` block matches existing `HAVE_TV` / `HAVE_HF` skip patterns in `device_compare_wattcomplete.py`.
- [ ] *Note for maintainer:* Written and syntax-checked on Linux. Since I don't have Apple Silicon silicon to execute the full benchmark battery, please verify the script run on macOS.

### DIFF
--- a/bench/device_compare_wattcomplete.py
+++ b/bench/device_compare_wattcomplete.py
@@ -578,7 +578,7 @@ def wl_vit_b16():
     if not HAVE_TV:
         note(wl, "skipped - torchvision unavailable.")
         return
-    # i very strongly do believe this is the right way to do it?
+    # graceful skip: examples/vit_demo.py is not always present (optional / gitignored); see #141
     try:
         sys.path.insert(0, str(REPO / "examples"))
         import vit_demo as vd
--- a/bench/device_compare_wattcomplete.py
+++ b/bench/device_compare_wattcomplete.py
@@ -578,8 +578,13 @@ def wl_vit_b16():
     if not HAVE_TV:
         note(wl, "skipped - torchvision unavailable.")
         return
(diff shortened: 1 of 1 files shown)


bench/device_compare_wattcomplete.py
```diff
@@ -578,8 +578,13 @@ def wl_vit_b16():
     if not HAVE_TV:
         note(wl, "skipped - torchvision unavailable.")
         return
-    sys.path.insert(0, str(REPO / "examples"))
-    import vit_demo as vd
+    # i very strongly do believe this is the right way to do it?
```
sbryngelson: Just this one line to tidy before merge: swap the dev-note for a comment that says *why* the guard is here. You can accept this suggestion directly:

```suggestion
    # graceful skip: examples/vit_demo.py is not always present (optional / gitignored); see #141
```

sbryngelson (commented):
Thanks for picking this up, @8nt0n -- and welcome. This is exactly what #141 asked for: guard the import and skip gracefully (mirroring the existing `if not HAVE_TV: note(...); return` idiom in this same function), rather than stubbing fake numbers. `ModuleNotFoundError` is an `ImportError` subclass, so an absent `examples/vit_demo.py` is caught, and `--perf-full` now completes with the ViT row skipped instead of aborting the whole battery.

Just one tiny cleanup suggested inline (the dev-note comment) -- accept that suggestion and this is good to merge. Nice, well-scoped first contribution. Closes #141.

8nt0n:
Thanks for the review and the warm welcome!

PR #161, opened 2026-08-03
compile: declare input activation ranges so opt=1 keeps int8 when provably safe

axiom-of-choice:
Implements #155, the piece you wanted landed before #156.

**This supersedes #156** rather than stacking on it: it carries that gate plus the bound machinery that turns it into an optimization instead of a near-blanket decline, so there is no regression window between the two. Close #156 in favour of this, or say the word and I will.

## Two pieces, per the issue

**1. Declare a range.** `af.input(shape, max_abs=6.0)` records a promise about `|value|` at runtime, stored on the node so `_node_max_abs` returns it instead of `None`. It is a promise, not a clamp, and the docstring says so. Undeclared inputs stay unbounded, so #153's fail-closed default is untouched.

**2. Propagate it.** `_propagate_max_abs` walks the graph conservatively. Anything not modelled returns `None` rather than a guess, since over-estimating only declines a lossy variant more often while under-estimating would wrongly keep one. Covered: non-expanding elementwise and shape ops, scalar `mul`/`add`, `add`/`sub`/`mul`/`minimum`/`maximum` over both operands, `clip` by its limits, the `matmul`/`linear` weight row-sum bound, `bmm`, and the normalizations when their affine is baked.

## The part worth more than the declaration

The saturating activations are bounded *whatever* feeds them, so they rescue an undeclared graph with no annotation:

```
input undeclared            bound=None       int8 declined
input undeclared -> softmax bound=1.0        int8 KEPT
input undeclared -> sigmoid bound=1.0        int8 KEPT
input max_abs=6.0           bound=6.0        int8 KEPT
input max_abs=6.0 -> relu   bound=6.0        int8 KEPT
input max_abs=1e5           bound=100000.0   int8 declined
```

So anything downstream of a `softmax`, `sigmoid`, `tanh`, `l2_norm` or `relu6` keeps the fast path for free. That covers a lot of real attention and classifier graphs without anyone touching their code, which is what makes this more than a fancier option 3.

## `_act_max_abs` changed meaning

In #156 it bounded the graph *output*. That was wrong for the purpose: what an int8-weight variant must encode is the activation feeding each weight-consuming node. It now walks to every `matmul`/`linear`/`bmm`/`conv` and bounds `srcs[0]` per node, returning `None` if any one of them is unbounded.

## Verification

Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2.

| | result |
|---|---|
| `bench/opt1_int8_saturation_repro.py` | **179/496 non-finite to 0/496** |
| declared `max_abs=8.0`, real peak 7.1 | int8 kept, finite, relerr **4.6e-3** (lossless path: 2.5e-4) |
| `tests/test_input_ranges.py` | 21 passed |
| `tests/test_opt1_act_ceiling.py` (from #156) | 10 passed, unchanged |
| `test_routes`, `test_tune_guards`, `test_shapes` | 47 passed together |
| `ruff` / `pyright aneforge` / pre-commit | clean (pyright's 9 are the pre-existing optional imports) |

The declared-range case is the one that needed on-device proof rather than a build-level assertion: the gate lets int8 through there, so it has to actually be correct, not just permitted. relerr 4.6e-3 against the fp32 reference is the expected int8 cost and sits inside `_ACCURACY_TOL`.

## Your review nit from #156

`bench/opt1_int8_saturation_repro.py` is committed here, since you said a committed on-device repro would be good to have, and the test docstring that referenced it while it was absent is fixed.

## Judgement calls worth flagging

**The `_NORM_SIGMA = 8.0` cap.** A normalized activation is O(1) but not provably bounded, since fp16 division by a small variance can overshoot, so bounding `layer_norm`/`rms_norm` output needs *some* assumption. I used `max|gamma| * 8 + max|beta|`, only when the affine is baked. 8 sigma is deliberately loose, but it is still an assumption rather than a proof, and it is the one part of this I would understand you rejecting. Dropping those four ops from `_BOUNDED_OUT` costs only bounds *through* a norm and changes nothing else.

**Untested combination.** I did not exercise `max_abs` together with `compress=` or a non-fp16 `dtype`; the declaration is inert in both paths as far as I can see, but I have not proven it.

### DIFF
--- a/aneforge/_compile.py
+++ b/aneforge/_compile.py
@@ -1094,7 +1094,7 @@ class CrossChipFP16Warning(UserWarning):
 
 
 def _node_max_abs(t: Tensor):
-  """Static bound on a node's value magnitude: a baked const's max, else None."""
+  """Static bound on a node's value magnitude: a baked const's max, a declared input bound, else None."""
   v = t.attrs.get("value")
   if v is not None:
     try:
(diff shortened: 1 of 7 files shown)


sbryngelson:
This is the right shape for #155, and the `_BOUNDED_OUT` insight is what makes it *not* a regression: softmax / sigmoid / tanh / l2_norm / relu6 bound the activation with no declaration, so anything downstream keeps int8 for free -- which covers most real graphs (every transformer has a softmax) without users touching their code. The propagation is appropriately conservative (norms via `gamma*8sigma + beta`, matmul via weight row-sums, over-estimation always safe, unmodelled -> None fail-closed), and `max_abs` as a promise-not-clamp with fail-closed default is exactly the contract. Test coverage is thorough, the repro is committed, and the docstring nit is fixed. It supersedes #156 -- I'm closing that in favor of this.

**One blocker before merge: this branch is behind #157 and would revert it.** Its `measure()` gate is back to `if relerr > tol` (head `_optimize.py:364`), where main now has `if not (relerr <= tol)` from #157 (merged). Merging as-is re-opens the exact `nan` fail-open #157 just closed. Please **rebase on current main** -- after that the diff should stop touching `measure()` entirely, since #161 doesn't intend to change it. (I'll hold CI approval until the rebase so it tests the right thing.)

Two smaller, non-blocking notes:
- The #157 follow-up still applies: the `nan` can also originate in `_measure_with_ref` (`:628`) and flow into `tune_precision`'s `min(usable, key=lambda r: r["relerr"])` fallbacks, which neither #157 nor this touches. Worth a guard eventually (drop non-finite relerr before the `min`), separate PR.
- `_NORM_SIGMA = 8` is the one place the bound is a heuristic rather than exact -- fine because over-estimation is always safe (and you comment it), just flagging it as the assumption to revisit if a real normalized activation ever exceeds it.

Issue #132, opened 2026-07-31
linalg: solve (general A x = b via LU)


sbryngelson:
A plain `solve(A, b)` is missing even though the pieces exist: `lu_pivoted` already factors on the ANE, so `solve` is forward-substitution `L y = P b` then back-substitution `U x = y`, i.e. two `solve_triangular` (#81) calls with the pivot applied host-side. Add it to `aneforge/linalg.py` as the direct (non-iterative) complement to `conjugate_gradient`/`gmres`, and accept a matrix right-hand side while you are there if `solve_triangular` grows matrix RHS (see #106).

Oracle is `np.linalg.solve` on well-conditioned square systems, style per `tests/test_linalg.py`. Note the fp16 conditioning limit and choose test matrices accordingly.

Verifying numerics needs any Apple Silicon Mac (the tests dispatch to the ANE); CI runs only the off-device subset, so state in the PR that the on-device tests passed locally.

Issue #103, opened 2026-07-22
linalg: matrix_power


sbryngelson:
Integer matrix powers by binary exponentiation over the module's `_ane_gemm` (square-and-multiply, so `A^n` costs O(log n) on-engine gemms). Handle n=0 (identity) and reject negative n (pair naturally with a future `inv`). Note fp16 error compounds with each squaring -- test at modest powers (say up to 8) on well-conditioned matrices with tolerances that acknowledge it.

Oracle is `np.linalg.matrix_power`; style per `tests/test_linalg.py`.

Verifying numerics needs any Apple Silicon Mac (the tests dispatch to the ANE); CI runs only the off-device subset, so state in the PR that the on-device tests passed locally.

PR #171, opened 2026-08-03
linalg: expm(A) by scaling and squaring

### LINKED ISSUE #133: linalg: expm (matrix exponential)
`expm(A)` via scaling-and-squaring: pick `s` so `||A / 2^s||` is small, approximate `exp(A / 2^s)` with a low-order Taylor or Pade sum (a handful of `_ane_gemm` products, exactly the building blocks `matrix_power` (#103) uses), then square the result `s` times. Everything is on-ANE matmul; the host only picks `s` from a norm (the new `norm`, #118).

Add to `aneforge/linalg.py`. Oracle is `scipy.linalg.expm` on well-conditioned, modest-norm matrices -- fp16 error compounds through the squarings, so document the regime the way `matrix_power` does and keep the test matrices tame. This is the meatier one of the batch (order selection + squaring loop), hence `help wanted`.

axiom-of-choice:
Fixes #133.

**Stacked on #170** (`solve`, #132), since both touch `linalg.py`. Merge that first and this becomes a two-file diff; happy to rebase onto `main` instead if you would rather take them in the other order.

## Approach

Exactly what the issue describes: the host picks `s` from the 1-norm so `A / 2^s` has norm at most 1/2, which is where a low-order Taylor sum is accurate, then the result is squared back `s` times. The Taylor sum and the squarings are all `_ane_gemm`, the same building block `matrix_power` (#103) uses, so the whole thing is O(order + s) on-engine gemms.

The term is built by recurrence, `term = term @ As / k`, so each order costs one gemm rather than recomputing a power, and `1/k!` folds into a host-side scalar so it needs no extra dispatch.

## Verification

Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2, against `scipy.linalg.expm`.

Modest norm, well conditioned:

| n | relerr |
|---|---|
| 4 | 4.8e-4 |
| 6 | 7.8e-4 |
| 8 | 7.6e-4 |

The scaling half is the part worth testing separately, since a Taylor sum alone would diverge there:

| 1-norm | s chosen | relerr |
|---|---|---|
| 0.64 | 1 | 5.2e-4 |
| 2.80 | 3 | 2.6e-3 |
| 6.54 | 4 | 2.2e-3 |

| Check | Result |
|---|---|
| `tests/test_linalg.py` | **74 passed** (57 on `main`, 9 from #170, 8 here) |
| `ruff` / `pyright` / `.githooks/pre-commit` | clean |

Two of the tests avoid a reference solver entirely, which matters because scipy is not in the dev extra (the scipy-based ones use `importorskip`): `expm(0)` must be the identity, and a diagonal argument must equal the elementwise `exp`, which pins the Taylor sum on its own.

## Judgement calls

**`order=8` as the default.** With the norm scaled to 1/2, the term at k=8 is already below fp16 resolution, so more orders buy nothing and cost gemms. It is a parameter rather than a constant in case you want it tunable, and `order < 1` rejects.

**The threshold is 1/2 rather than a Pade-style table.** scipy switches order and threshold together from a lookup table tuned for fp64; at fp16 the accuracy is dominated by the squaring error, not the truncation, so a single conservative threshold is the honest choice. The docstring says so, and says this is for modest norms on well-conditioned input, the same caveat `matrix_power` carries.

I used the 1-norm computed host-side rather than calling `norm(A, 1)` (#118), since that dispatches a graph to get a scalar the host then needs immediately; happy to switch if you would rather have the one code path.

### DIFF
--- a/aneforge/linalg.py
+++ b/aneforge/linalg.py
@@ -531,7 +531,7 @@ def eigvals(A, iters: int = 60):
 __all__ = [
   "conjugate_gradient", "jacobi", "gauss_seidel", "iterative_refine",
   "least_squares", "lsqr", "gmres",
-  "qr", "cholesky", "lu", "lu_pivoted", "solve", "solve_triangular",
+  "qr", "cholesky", "lu", "lu_pivoted", "solve", "solve_triangular", "expm",
   "eigh", "eigvals", "generalized_eigh", "dominant_eig",
   "svd", "dominant_svd", "svdvals_topk", "randomized_svd", "pca",
 ]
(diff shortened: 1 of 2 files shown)


sbryngelson (approved):
Reviewed and it's a clean approve on the expm code -- the scaling picks `s` from the 1-norm so the scaled matrix lands at <= 1/2 (verified against your table: 6.54 -> s=4 -> 0.41), the `term = term @ As / k` recurrence builds `exp(As)` at one gemm per order, and the `s` squarings recover `exp(A)`; zero matrix falls through to the identity. The two reference-free tests (identity, diagonal closed form) are exactly the right pins given scipy isn't in the dev extra. Nice.

#170 just merged, so please take the rebase-onto-main option you offered -- `git rebase --onto main feat-132-linalg-solve feat-133-linalg-expm` should drop the now-redundant solve commits and leave the expm-only two-file diff. Once it shows just `expm` I'll merge (closes #133). Thanks for the clean, well-verified series.

axiom-of-choice:
> Reviewed and it's a clean approve on the expm code -- the scaling picks `s` from the 1-norm so the scaled matrix lands at <= 1/2 (verified against your table: 6.54 -> s=4 -> 0.41), the `term = term @ As / k` recurrence builds `exp(As)` at one gemm per order, and the `s` squarings recover `exp(A)`; zero matrix falls through to the identity. The two reference-free tests (identity, diagonal closed form) are exactly the right pins given scipy isn't in the dev extra. Nice.
> 
> #170 just merged, so please take the rebase-onto-main option you offered -- `git rebase --onto main feat-132-linalg-solve feat-133-linalg-expm` should drop the now-redundant solve commits and leave the expm-only two-file diff. Once it shows just `expm` I'll merge (closes #133). Thanks for the clean, well-verified series.

No Worries. Just rebased and pushed

Issue #164, opened 2026-08-03
bench: contribute a numeric_cliffs.py datapoint for your chip


sbryngelson:
Companion to the #137 roofline drive, but for **correctness ceilings** instead of throughput. `bench/numeric_cliffs.py` measures the magnitudes at which fp16 math on the engine silently returns the wrong answer -- the three cliffs ANEForge has characterized (see the module docstring and `docs/cross-chip.md`):

1. **matmul_saturation** -- a matmul output flips finite->inf near `fp16_max/2` (~32752), for every contraction size K (#115).
2. **slice_saturation** -- a width-offset slice routing through the pre-A16 Q.4 x16 crop-DMA that clamps `|value| > 4094` to +/-inf. Documented EXACT on A16/M5, so a modern part should report **no** cliff.
3. **reduce_exactness** -- a reshape->reduce_sum of small integers stops being bit-exact crossing the fp16 integer-spacing boundary at 2048.

**Each cliff is potentially silicon-dependent**, and that is exactly the cross-hardware question the roofline work asks: does your chip's matmul cliff sit where the M2/M5 cliff does? Is the slice cliff present (pre-A16) or gone (A16/M5)?

**How to contribute (2-minute run, you get credited):**
```sh
PYTHONPATH=. python3 bench/numeric_cliffs.py
```
Paste the JSON it prints as a comment here, along with your chip (e.g. "Apple M3 Pro, 18-core GPU"). No code required. If you want, `bench/_machine.py` (`python3 -c "import bench._machine as m, json; print(json.dumps(m.fingerprint()))"`) prints a machine fingerprint you can paste alongside.

Good first issue for anyone with an Apple Silicon Mac -- the more chip generations we get, the sharper the cross-silicon cliff map.

PR #180, opened 2026-08-04
docs: add numeric correctness cliffs reference

### LINKED ISSUE #178: docs: a numeric cliffs reference page
ANEForge has characterized three per-silicon **fp16 correctness ceilings** -- magnitudes at which the engine silently returns the wrong answer -- but they are scattered across issues and `docs/cross-chip.md`. Consolidate them into one reference page under `docs/` (add it to the `mkdocs.yml` nav).

**The three cliffs to document** (source of truth: `bench/numeric_cliffs.py`'s module docstring and the issues below):

1. **matmul saturation** -- a matmul output flips finite -> inf near `fp16_max/2` (~32752), for every contraction size K. Issue #115.
2. **slice Q.4 x16 crop-DMA clamp** -- a width-offset slice routes through a fixed-point path that clamps `|value| > 4094` to +/-inf on pre-A16 parts; documented EXACT on A16/M5. See `docs/cross-chip.md`.
3. **reduce exactness** -- a reshape -> reduce_sum of small integers stops being bit-exact crossing the fp16 integer-spacing boundary at 2048 (a row summing to 2048 reads 2050).

**The page should:** state each cliff, why it happens (fp16 representable range / fixed-point datapath / integer spacing), which silicon it applies to, and how a user avoids it. Cross-link to `bench/numeric_cliffs.py` and the datapoint drive in #164 so readers can measure their own chip.

**No Apple Silicon required** -- this is a docs-only change (verify the mkdocs nav builds with `mkdocs build`, which CI also checks).

BortnikMaxim:
## Summary

- consolidate the three characterized fp16 correctness cliffs in one reference page
- explain each cause, measured silicon scope, failure mode, and practical avoidance
- link the measurement script, source issue, and datapoint drive
- add the page to the MkDocs navigation

## Testing

- `.venv/bin/mkdocs build --strict`

Closes #178

### DIFF
--- a/docs/numeric-cliffs.md
+++ b/docs/numeric-cliffs.md
@@ -1,0 +1,79 @@
+# Numeric correctness cliffs
+
+ANE programs use fp16, but the largest finite fp16 value is not a guarantee that
+every operator can produce or carry every value up to that limit. Some hardware
+paths have lower **correctness cliffs**: a mathematically finite result becomes
+infinite, or an integer result is rounded. These are correctness limits rather
+than performance limits.
+
+The values below describe the untouched lowering (`opt=0`). They have been
+measured on M2 Pro (A14 family) and M5 (A16 family); run the sweep on other chips
+instead of assuming an unmeasured family behaves identically.
+
+| Operation | Cliff | Observed silicon | Failure |
+| --- | --- | --- | --- |
+| `matmul` | output magnitude around 32752 | M2 Pro and M5 | finite output becomes `+/-inf` |
+| width-offset slice | input magnitude above 4094 | pre-A16; absent on A16/M5 | sliced value becomes `+/-inf` |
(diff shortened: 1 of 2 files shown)


sbryngelson (approved):
This is excellent, @BortnikMaxim -- and welcome! I checked every number against the source and the codebase and they're all right: matmul at `fp16_max/2` (~32752), the slice clamp at `65504/16 = 4094` (pre-A16, exact on A16/M5), the reduce boundary at 2048, and the family labels match `_targets.py` (M2 Pro = A14 / H14, M5 = A16). The avoidance guidance and the cross-links to `bench/numeric_cliffs.py` and the datapoint drive (#164) are exactly what the issue wanted. Merging as soon as the docs build goes green. Thank you!

Issue #181, opened 2026-08-04
bench: tiled vocab head so the decode roofline measures on every ANE family (+ re-run landed machines)


sbryngelson:
The headline **Decode @b1 (tok/s)** roofline is measurable only on A16+ (M5) today, and reads `n/a` on every older family (M1/M2/M1 Max). Root cause: `bench/decode_measurement.py` builds the vocab head as a single untiled matmul `h @ Wvocab.T -> [B, 32000]`, and 32000 exceeds the 16384 max matmul dimension on the A13-A15 families, which raises `NotImplementedError` for the ANE decode row. Only A16 (max-dim 65536) fits it untiled.

**Fix (in progress):** tile the vocab head along the output axis so no single matmul output dim exceeds 16384 (a 32000 vocab -> two chunks of 16384 + 15616). Tiling the same way on *every* chip keeps the number comparable across generations. Verified on M5: the tiled result is bit-identical to the untiled one (relerr 0.00 vs untiled, same fp16 accuracy vs numpy), so it changes nothing except making the head fit the cap.

**What remains after the code lands (this issue tracks it):** the committed datapoints predate the tiled head, so they need a re-run to populate decode on the older families:

- [ ] Apple M1 (@diegobauavi) -- re-run `bench/roofline_suite.py --perf` and update the JSON
- [ ] Apple M1 Max (@sbryngelson)
- [x] Apple M2 Pro (@axiom-of-choice)
- [x] Apple M5 Pro (@sbryngelson) -- re-run so its number is the tiled one too, for apples-to-apples

Until then, **Peak fp16 GEMM (TF/s)** is the headline performance metric (measured on every machine, no dim-cap issue), and the decode column is footnoted as migrating.

Related: the `roofline_analysis.py` `KeyError: 'ane'` crash visible in the older JSONs is already fixed by #147; those JSONs are just stale, and a re-run picks up the fix.

sbryngelson:
@diegobauavi (M1) and @axiom-of-choice (M2 Pro) -- would you be up for re-running the perf suite so the decode column fills in for your chips?

The tiled vocab head is in #182. I verified on an M5 that the tiled result is bit-identical to the untiled one, but I can't confirm the actual untiled->tiled flip without an A13-A15 part -- so a re-run on your hardware would both **populate your decode number** and **verify the fix** on the families it's actually for. That's the real test.

Two ways to help, whichever is easier:

- **Now (also verifies #182):** check out that branch and run
  ```sh
[...]

PR #183, opened 2026-08-04
bench: return lm_head tiles as output ports so decode measures on family 3

axiom-of-choice:
Refs #181. Follow-up to #182: the tiled head still doesn't measure on family 3, because the tiles were rejoined with `af.concat` before compile.

`_retarget_for` checks `rep.oversize` over **every tensor in the graph**, not just matmul outputs (`aneforge/_compile.py:1280`), so the concat is itself over the cap:

```
NotImplementedError: aneforge: tensor (1, 32000) (op 'concat') exceeds ANE
family 3's max dimension 16384; tile the graph or target a higher chip.
```

The tiling was right — the chunks come out `(1, 16384)` and `(1, 15616)`, both under the cap — but reassembling them in-graph put the result back over it, so decode still read `n/a` on M2 Pro. It passes on A16+ only because that cap is 65536, which is why it looked verified on M5.

This keeps the tiles as separate output ports via `compile_multi`. `logits` is the terminal node, so nothing downstream needs them fused on-engine; they're stitched host-side for the relerr check only, **outside the timed region**. A vocab that fits in one tile still takes the plain `af.compile` path, so the single-output program is unchanged.

Measured on Apple M2 Pro (family 3, `max_tensor_dim` 16384), `--quick`:

| device | before | after | relerr |
| --- | --- | --- | --- |
| ANE B=1 | `NotImplementedError` | **277.4 tok/s** | 1.65e-2 |
| ANE8 B=1 | `NotImplementedError` | **505.4 tok/s** | 3.90e-2 |
| ANE B=32 | `NotImplementedError` | **3685.1 tok/s** | 1.67e-2 |
| ANE8 B=32 | `NotImplementedError` | **4180.8 tok/s** | 3.51e-2 |

CPU is unchanged at 46.2 / 808.4 tok/s, so ANE is 6.0x CPU at B=1 and 4.6x at B=32.

Two things worth your call, both of which affect whether the number is comparable across generations (which is the point of #181):

1. **Where the concat belongs.** On A16+ the untiled graph produces `[B, vocab]` on-engine; here the stitch is host-side and outside the timed region. If you'd rather the tok/s include it everywhere, the timing loop is the place to change, not this.
2. **The fp16 relerr.** ~1.7e-2 fp16 / ~3.5e-2 int8 across a full decode stack looks plausible to me, but I can't check it against the untiled path — that needs an A16+, which I don't have. Worth confirming the tiled and untiled heads agree on M5 before trusting the cross-family comparison.

I have the M2 Pro `--perf` datapoint ready to submit separately once this lands, so the JSON carries the decode numbers rather than another `n/a` row. Holding it back for that reason.

### DIFF
--- a/bench/decode_measurement.py
+++ b/bench/decode_measurement.py
@@ -28,6 +28,7 @@ relerr = dc.relerr
 
 if HAVE_ANE:
     import aneforge as af
+    from aneforge._compile import compile_multi   # not re-exported; tests use it the same way
 if HAVE_MLX:
     import mlx.core as mx
 
@@ -135,12 +136,18 @@ LMHEAD_TILE = 16384
 
 
 def _lm_head(h, Wvocab_f16):
-    """logits = h @ Wvocab.T -> [B, vocab], tiled along vocab so each matmul output dim <= LMHEAD_TILE."""
+    """logits = h @ Wvocab.T, tiled along vocab so each matmul output dim <= LMHEAD_TILE.
+
+    Returns a LIST of tiles, never a concatenated [B, vocab] tensor: the target gate rejects any
+    tensor over the family cap, not just matmul outputs, so an in-graph concat of the tiles is
+    itself oversize on family 3 (a 32000 vocab concat is [B, 32000] > 16384) and the tiling buys
+    nothing. Keeping the tiles as separate output ports is what actually fits, and since logits is
+    the terminal node nothing downstream needs them fused on-engine.
+    """
     V = Wvocab_f16.shape[0]
     if V <= LMHEAD_TILE:
-        return h.linear(Wvocab_f16)
-    parts = [h.linear(Wvocab_f16[i:i + LMHEAD_TILE]) for i in range(0, V, LMHEAD_TILE)]
-    return af.concat(parts, axis=1)                                # [B, sum(chunks)] = [B, vocab]
+        return [h.linear(Wvocab_f16)]
+    return [h.linear(Wvocab_f16[i:i + LMHEAD_TILE]) for i in range(0, V, LMHEAD_TILE)]
 
 
 # ANE graph (aneforge) - full per-token stack, batch B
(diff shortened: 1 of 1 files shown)


axiom-of-choice:
Ran the full `roofline_suite.py --perf` on this branch (all four steps rc=0, sudo granted) so the decode row is measured through the real suite rather than the script alone. Apple M2 Pro, family 3, tree clean at the branch head.

| device | b=1 tok/s | b=1 tok/s/W | b=32 tok/s | b=32 tok/s/W | relerr |
| --- | --- | --- | --- | --- | --- |
| CPU | 41.2 | 9.6 | 822.0 | 151.6 | 5.4e-7 |
| ANE | **281.7** | **96.4** | **3746.4** | **2275.0** | 1.65e-2 |
| ANE8 | **493.2** | 57.9 | **4256.5** | n/a | 3.51e-2 |

All three ANE/ANE8 rows read `NotImplementedError` on `main`. The tok/s reproduce the `--quick` numbers in the description to ~1.5% (281.7 vs 277.4, 3746 vs 3685), measured in a separate process, so the result isn't a one-off.

The `--perf` run also fills the decode watts column, which no M2 Pro submission has had: ANE decode is 96.4 tok/s/W at b=1 versus CPU's 9.6, a 10x energy advantage that was previously unmeasurable on this family.

I'm not attaching this report as a datapoint since it records `branch: fix-181-lmhead-tiles-as-ports`; I'll re-run against `main` and submit that separately once this lands.

One caveat I can't close myself: the fp16 relerr of ~1.65e-2 (int8 ~3.5e-2) is measured against the fp64 reference, not against the untiled head, so it doesn't prove the tiled and untiled paths agree. Confirming that needs an A16+, where the untiled graph fits. Worth a check on M5 before the numbers are compared across families.

sbryngelson (approved):
Great catch, and thank you -- this is a real bug in #182. My tiling fixed the matmul outputs but rejoined them with `af.concat`, putting the result back to `[B, 32000]`; since the target gate checks every tensor (not just matmul outputs), the concat was itself oversize on family 3. It only passed on M5 because A16's cap is 65536, so I verified it exactly where the bug was hidden -- the A16 blind spot you flagged. Keeping the tiles as output ports and stitching host-side is the right fix.

**On your question 2 (the one that needs an A16+): I ran it on the M5.** Untiled `[1,32000]` head vs your tiled output-ports head, stitched via `output_ports`:

```
max|delta| = 0.000e+00   relerr = 0.000e+00   argmax match: True
```

Bit-identical. The tiling only partitions the vocab columns and the stitch reassembles them losslessly, so the two paths agree exactly. That ~1.7e-2 you measured is fp16-vs-fp64 on the full stack -- the untiled head shows the same. So the cross-family number is comparable, which was the point of #181.

**On question 1 (where the stitch belongs):** agree with your choice -- keep it host-side, outside the timed region. It's bit-identical so it changes nothing numerically, and real decode only needs the argmax/sampling, not the full vocab vector fused on-engine. The timed region measures the on-engine tile matmuls on every family, which is the quantity that should be comparable. No change.

One consequence worth noting: with this, M5 (vocab 32000 > 16384) now also takes the 2-tile path, so every family runs the same graph shape -- good for comparability. Its committed 117 tok/s was the untiled number, so M5 wants a re-run too (already on the #181 checklist).

On the bandwidth/ridge code-path point from #181: agreed, and good catch -- the pre-#147 M1/M2 JSONs use the raw-sweep fallback while M1 Max/M5 used the synthesis path, so part of that column's spread is code path, not silicon. A re-run on current `main` fixes it, so your M2 Pro submission will be consistent.

Merging now. Submit the M2 Pro `--perf` datapoint against `main` and I'll tick the box on #181.

PR #185, opened 2026-08-05
bench: M2 Pro roofline datapoint with the tiled decode head

axiom-of-choice:
Closes the `Apple M2 Pro` checkbox in #181. Re-run of `roofline_suite.py --perf` on `main` now that #183 landed, so decode populates instead of reading `n/a`. All four steps rc=0, sudo granted, AC power.

## Summary

| | committed (Aug 01) | this run |
| --- | --- | --- |
| Decode @b1 (tok/s) | `n/a` | **276** |
| Peak fp16 GEMM (TF/s) | 3.3 | 3.4 |
| Bandwidth (GB/s) | 0.96 | 0.98 |
| Ridge (FLOP/byte) | 3404 | 3429 |
| Peak perf/W (GF/s/W) | 1379 | 744 |
| `roofline_analysis.py` | rc=1 | rc=0 |
| `idle_pkg` | 10973 mW | 368 mW |

Decode per device, b=1 / b=32 tok/s, and tok/s/W at b=1:

| device | b=1 | b=32 | b=1 tok/s/W | relerr |
| --- | --- | --- | --- | --- |
| CPU | 62.4 | 478.0 | 1.3 | 9.7e-7 |
| GPU | 450.5 | 7156.0 | 99.3 | 2.4e-3 |
| ANE | **275.8** | **3332.9** | **94.8** | 1.65e-2 |
| ANE8 | **492.5** | **4009.8** | **148.8** | 3.90e-2 |

ANE decode is 4.4x CPU at b=1 and 73x its tok/s/W. The 94.8 tok/s/W reproduces the 96.4 I reported in #181 from the #183 branch, measured in a separate process on a quieter machine.

## Two things a reviewer should read carefully

**Peak perf/W 1379 -> 744 is a correction, not a regression.** At the same N=4096 this run measures *more* throughput (3365 vs 3263 GFLOP/s) and *more* power: 4.30 W active package vs 2.14 W, ANE rail 2610 vs 1682 mW. The old submission ran with `idle_pkg` at 10973 mW, so subtracting idle from active erased most of the real draw and left an understated denominator. This run idles at 368 mW, and CV at that point is 8.1% vs 14.8% before. The old 1379 was the inflated number.

**`roofline_analysis.py` going rc=1 -> rc=0 changes where two columns come from.** The committed M1 and M2 Pro JSONs predate #147, so `_perf_headline()` fell back to the raw sweep peak for bandwidth/ridge while M1 Max and M5 Pro used the synthesis path. Part of the M1-vs-M1 Max gap in those columns is therefore code path, not hardware. This fixes the M2 Pro row; the M1 row still needs its re-run.

## Note on the `(dirty)` flag

The Machines table shows `64108472d01f (dirty)` for this row. That is the suite fingerprinting itself, not local edits: `bench/results/*_results.json` are tracked, `--perf` overwrites all five as it runs, and `_machine.py` reads `git status --porcelain` afterwards. Any `--perf` from a clean checkout marks itself dirty, which is why the M5 Pro row shows it too. The merge-base is exact (0 ahead / 0 behind) and no tracked file differs from `main` in this PR.

Worth fixing separately, since the flag currently cannot distinguish "contributor had uncommitted edits" from "the suite wrote its own outputs" — either gitignore those five files or exclude them from the dirty check in `_machine.py`. Happy to open that as its own PR; I left it out of here to keep the datapoint clean.

## Evaluation

Reproducibility: the b=1 tok/s reproduce across three separate `--perf` invocations on this machine (275.8 / 278.3 / 281.7 for ANE, 492.5 / 492.7 / 493.2 for ANE8), the spread tracking how quiet the machine was. The run committed here is the quietest of the three by idle power and has the lowest CV on every decode row; it is also the only one where no row carries the `loaded-package CV >35%` low-confidence flag.

`sustained/peak` for ANE b=32 is 0.90 here (0.50 in a noisier run), and CPU b=1 is 0.35 — the CPU rows are the least stable in this suite, which the existing `only 8 pm samples` flag already annotates.

## Risks

- fp16 relerr ~1.65e-2 (int8 ~3.9e-2) is measured against the fp64 reference, not against the untiled head, so it still does not prove the tiled and untiled paths agree. That check needs an A16+. Unchanged from #183 and I cannot close it myself.
- Numbers are `--quick` decode config (d_model 1024, 4 layers, vocab 32000), not `--perf-full`, matching how the other committed datapoints were produced.
- Single machine, single run committed. Immutable per the submissions policy, so re-measurement means a new file rather than an edit.

## Pre-merge

- CI `rooflines table in sync` should pass; `python bench/aggregate_rooflines.py --check` is clean locally (`ROOFLINES.md up to date (5 submission(s))`).
- Nothing else to do; the JSON and the regenerated table are the whole change.

## Post-merge

- #181 still needs the M1 (@diegobauavi) and M5 Pro re-runs; M1 Max too if you want it on the tiled head. Once M1 lands, both pre-#147 rows are off the fallback path and the bandwidth/ridge column is comparable across the table.
- The decode footnote under the table can drop once no row reads `n/a`.


sbryngelson (approved):
Clean, and the write-up is exactly what a data submission should be. The M2 Pro decode number (276 tok/s, 4.4x CPU) is the tiled head from #183 working on real A14/A15 hardware -- thanks for closing that loop.

Both corrections make sense and I'd rather have them:
- The perf/W 1379 -> 744 is well-diagnosed: the old idle_pkg at 10973 mW erased most of the real draw from the denominator. 744 at a 368 mW idle with lower CV is the honest number.
- roofline_analysis going rc=0 fixes the bandwidth/ridge code-path inconsistency you flagged in #181; M1 still needs the same re-run.

Good catch on the `(dirty)` flag fingerprinting itself -- please do open that as its own PR (gitignore the five `*_results.json` or exclude them from the porcelain check in `_machine.py`); agreed it shouldn't ride along here.

Merging. That's the M2 Pro box in #181 done.

PR #188, opened 2026-08-05
bench: dirty flag ignores the suite's own outputs

axiom-of-choice:
Follow-up to #185, as you asked. Every `--perf` submission from a pristine checkout self-reported `dirty: true` — the M2 Pro row in #185 and the M5 Pro row before it.

## Cause

Each perf step writes its raw output to a **tracked** file under `bench/results/` (the `PERF_FAST`/`PERF_FULL` tables in `roofline_suite.py`), and the next step reads it back as input: `roofline_analysis.py` consumes `device_saturation_sweep_results.json`. By the time `_machine.py` takes the fingerprint, `git status --porcelain` reports four modified tracked files, and the flag fires on *the benchmark having run* rather than on the contributor's tree.

That defeats the field: it could not distinguish "the contributor had uncommitted edits" from "the suite wrote its own outputs", which is exactly the drift the fingerprint exists to make explicit.

Reproduced on a fresh clone, before this change:

```
$ git clone … && git status --porcelain      # clean
$ # simulate what a perf step does: rewrite its own result_file
$ git status --porcelain
 M bench/results/decode_measurement_results.json
$ python3 -c "import _machine; print(_machine._git_info()['dirty'])"
True
```

After: `False`, and `True` again as soon as a real source file is edited.

## Approach

`dirty` now excludes `bench/results/*_results.json`.

Kept as an exclusion in the dirty check rather than a `.gitignore`, of the two options you offered: those files are committed with real content (21–44 KB each) and are read as input by a later step, so untracking them would break the pipeline — and gitignoring an already-tracked file does not untrack it anyway.

Two details worth a look:

- The glob is matched with `PurePosixPath.match` plus a component-count check, not `fnmatch`. `fnmatch` lets `*` cross `/`, so `bench/results/*_results.json` would also swallow `bench/results/rooflines/*.json` — the committed submissions, which must stay visible as real changes. Covered by a test.
- `_git()` strips the whole stdout, so a leading `' M'` (modified-unstaged, the common case) arrives as `'M'`. Slicing `line[3:]` therefore ate a character off the **first** path, turning `bench/…` into `ench/…` and silently failing to match — which would have defeated the fix in precisely the case it targets. `_porcelain_paths` splits on whitespace instead. My own end-to-end check caught this after the unit tests passed; there is now a test pinning it.

## Tests

`tests/test_machine_dirty_flag.py`, 16 cases, pure string logic and no device:

- the glob: every `result_file` in both perf tables matches; committed submissions, `ROOFLINES.md`, source files, and `bench/results/<subdir>/*_results.json` do not
- the porcelain parser: renames (destination scored), quoted paths with spaces, staged/index states, blank lines, and the stripped-first-line case
- the composed decision: suite outputs alone → not dirty; a source edit alone or alongside suite outputs → dirty; a new submission JSON → dirty

```
$ python3 -m pytest tests/test_machine_dirty_flag.py -q
16 passed
```

`python bench/aggregate_rooflines.py --check` is clean, and `PYTHONPATH=. python3 bench/_machine.py` still emits a valid fingerprint (rc=0).

## Risks

- Behaviour change in a recorded field: submissions made after this read `dirty: false` where they previously read `true`. Already-committed JSONs are untouched, so the M2 Pro and M5 Pro rows keep their `(dirty)` label until those machines are re-run. If you would rather the table not mix the two conventions, a note under the Machines table would cover it — say the word and I will add it.
- The exclusion is a hardcoded glob, not derived from `roofline_suite.py`'s tables. A future perf step writing a tracked file that does not end in `_results.json` would re-introduce the flag. Deriving the list at import time would need `_machine.py` to import `roofline_suite.py`, which is the wrong direction (the suite imports `_machine`), so a test asserts every current `result_file` matches instead — that fails loudly if a new step drifts.
- No change to what a genuinely dirty tree reports.

## Pre-merge

Nothing. No new deps, no device needed for the tests.

## Post-merge

Unrelated to this but still open in #181: the M1 (@diegobauavi), M5 Pro and M1 Max re-runs. Those datapoints will land with the corrected flag.


sbryngelson (approved):
Clean fix, and the two details you flagged are the sharp ones. The `PurePosixPath.match` + component-count guard is right -- `fnmatch` would have let `bench/results/*_results.json` swallow `bench/results/rooflines/*.json`, which are the committed submissions that must stay visible. And the strip()/`line[3:]` interaction eating the first path's leading char is exactly the kind of thing that passes unit tests and then silently no-ops the fix in the field; good that you caught it end-to-end and pinned it with a test. 16 off-device cases covering the glob, the porcelain parser, and the composed decision is thorough. Merging -- thanks for closing this out.

PR #196, opened 2026-08-07
special: erf as its own polynomial, not 1 - erfc

### LINKED ISSUE #166: special: erf (companion to the existing erfc)
`aneforge/special.py` implements `erfc` (line ~57) but not `erf`. Add `erf`, the natural companion:

```python
def erf(x: Tensor) -> Tensor:
    return _const(x, 1.0) - erfc(x)   # erf(x) = 1 - erfc(x)
```

(or share the underlying polynomial directly if that is cleaner given how `erfc` is built -- follow the existing `_horner` / `_poly_in` style in the file).

**Oracle:** `scipy.special.erf`. Add a case to the module self-test (`special.py`'s `_selftest`, matching how `erfc`/`expm1`/`log1p` are checked) over a spread of x in roughly `[-4, 4]`.

**Testing:** the numeric check dispatches, so run it on any Apple Silicon Mac and note in the PR that it passed locally; CI runs only the off-device/build subset.

axiom-of-choice:
Fixes #166.

## Summary

The issue suggested `erf(x) = 1 - erfc(x)`. That reintroduces exactly the cancellation `erfc` exists to avoid, only mirrored: near 0 `erf` is small while `erfc -> 1`, so the subtraction loses the value being computed. `special.py:52` already makes the same point in the other direction ("direct, not 1-erf which cancels").

So `erf` gets its own polynomial: a deg-5 minimax of `erf(x)/x` in `x^2` on [0, 2], the same `x*poly(x^2)` shape as `sin`. It is odd, so it holds for negative `x` unchanged. Past |x| ~ 2 the roles swap and `1 - erfc(|x|)` is the accurate path, since `erfc` is small there; that is documented on the function and in `docs/developer/numerics.md` next to the existing `erfc` note.

## Evaluation

Measured on device over `geomspace(1e-3, 1)`, per-point relative error (the max-normalized `relerr` in the self-test harness divides by `|ref|.max()` ~ 1, which would hide precisely the small-x error this is about):

| form | per-point relerr |
|---|---|
| `x * P(x^2)` | **0.10%** |
| `1 - erfc(x)` | **187%** |

At `x = 0` the naive form returns `-0.000977`, a negative value for a function that is odd through the origin. The direct form returns exactly `0.0`.

Sanity check on the fit: the leading coefficient is `1.128338`, against `2/sqrt(pi) = 1.128379`, which is the `erf(x) = 2/sqrt(pi) x + O(x^3)` limit.

Module self-test on **Apple M2 Pro (Mac mini, macOS 26.5.2)**:

```
erf              [1e-3, 1] geom  relerr  1.05e-03  direct x*poly, vs 187% for 1-erfc(x) on the same grid
erf (odd/zero)   [-0.5, 0.5]     abserr  0.00e+00  sign-symmetric and exactly 0 at the origin
```

`1.05e-03` against an fp16 eps of `9.8e-04`, so this is rounding-limited rather than coefficient-limited, consistent with the other functions in the module.

## Tests

Six new cases, split across two files deliberately:

- `tests/test_special_trig.py` (3, on device): matches `scipy.special.erf` over [-2, 2]; accurate near zero where `1 - erfc` is not, asserting both that the direct form is under 5e-3 and that the naive form exceeds 100% on the same grid; odd and exactly zero at the origin.
- `tests/test_special_erf_form.py` (3, **new file, off device**): `erf`'s body does not call `erfc`, the leading coefficient recovers `2/sqrt(pi)`, and `erf` is exported. These are in a separate module because `test_special_trig.py` is `pytestmark = requires_ane`, so CI would skip the whole file.

All four numeric and form assertions fail if `erf` is written as `1 - erfc` (verified by substituting the naive version).

## Verification

- Corpus **GATE: GREEN 90/90**.
- `pytest tests/`: **860 passed, 4 skipped**. 6 failures are pre-existing and caused by optional deps not installed here (`transformers`, `torchvision`); confirmed identical on a clean `origin/main` worktree.
- `ruff` clean, `.githooks/pre-commit` clean, `pyright aneforge` reports 0 errors in `special.py` (the 12 in `onnx.py`/`_gguf.py` are pre-existing and unchanged).
- Per CONTRIBUTING: CI cannot reach an ANE, so the on-device tests above were run locally.

## Risks

Low. `erf` is a new public function, so nothing existing changes behavior; `erfc` and the shared helpers are untouched. The one judgement call is the [0, 2] range: beyond it the polynomial degrades (deg-5 on [0, 2] is 2.8e-4 in f64, well under fp16 noise, but a wider fit would cost accuracy where it matters), which is why the docstring points at `1 - erfc(|x|)` for larger arguments instead of widening the fit.

## Pre-merge / post-merge

None. No schema, build, or API changes; `erf` is additive and `__all__` is updated.

### DIFF
--- a/aneforge/special.py
+++ b/aneforge/special.py
@@ -61,6 +61,20 @@ def erfc(x: Tensor) -> Tensor:
   return poly * (x * x * -1.0).exp()
 
 
+# erf: direct deg-5 minimax of erf(x)/x in x^2 on [0, 2], NOT 1 - erfc(x).
+# erf is small near 0 while erfc -> 1, so `1 - erfc` cancels exactly where erf is
+# wanted: in fp16 it gives 22% relative error at x = 0.01 and returns -0.000977
+# at x = 0 (a negative value for a function that is odd through the origin).
+# Same x*poly(x^2) shape as `sin`; leading coefficient recovers 2/sqrt(pi).
+_ERF_P = [1.128338362755043, -0.3752941986401827, 0.1100666568321253,
+          -0.02340602798434916, 0.003153058641086897, -0.0001952199570501452]
+
+
+def erf(x: Tensor) -> Tensor:
+  """Error function erf(x) for |x| <= 2, as x*poly(x^2) (deg-5 minimax of erf(x)/x); odd, so it holds for negative x. Past |x| ~ 2 use 1 - erfc(|x|), where erfc is small and the subtraction no longer cancels."""
+  return x * _poly_in(x * x, _ERF_P)
+
+
 # expm1 / log1p - the small-argument cancellation pair
 
 def expm1(x: Tensor) -> Tensor:
@@ -157,7 +171,7 @@ def log_wide(x: Tensor, sqrts: int = 3) -> Tensor:
 
 
 __all__ = [
-    "sin", "cos", "erfc", "expm1", "log1p", "gamma", "lgamma", "gamma_via_lgamma",
+    "sin", "cos", "erf", "erfc", "expm1", "log1p", "gamma", "lgamma", "gamma_via_lgamma",
     "bessel_j0", "bessel_i0", "bessel_k0", "exp_wide", "log_wide",
 ]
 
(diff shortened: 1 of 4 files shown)


sbryngelson (approved):
Nice -- this is better than the issue asked for, and you were right to reject the `1 - erfc` premise. The mirror-cancellation argument is exactly right (187% vs 0.10%, and a negative value at an odd function's origin is a great tell), and catching that the harness's max-normalized relerr would hide the small-x error is the sharp part -- per-point relerr is the correct metric here. The source-inspection guard test is a clever way to keep CI protecting the formulation without an ANE.

One thing I'll note rather than block on: `erf` is accurate on [0, 2] and the |x|>2 path is left to the caller (documented), so `erf(3.0)` extrapolates the polynomial and returns garbage rather than ~1. `erfc` is the right tool for the tail, but a Gaussian-CDF caller passing a z-score > 2 would hit it. A `select`-based `|x|>2 -> sign(x)*(1 - erfc(|x|))` branch would make it robust across the whole line -- worth a small follow-up, not needed here.

Merging. Thanks -- clean fix for #166.

PR #198, opened 2026-08-10
bench: roofline run M5 10 core

danlee2002:
Ran benchmark script twice with `--perf` flag.


sbryngelson:
Welcome, @danlee2002, and thanks -- this is the first **base M5** on the board (decode 228 tok/s, a nice cross-check against the M5 Pro). 🎉

I pushed a small merge to your branch: another datapoint (M4) landed just before yours, so the regenerated `ROOFLINES.md` / `rooflines.json` conflicted. I merged `main` in and re-ran `python3 bench/aggregate_rooflines.py` so the tables now include every machine -- nothing changed in your actual measurement, just the generated aggregates.

One thing to double-check when you get a chance: the fingerprint auto-detected **8 GPU cores** (`gpu_cores: 8`), but the PR title says 10-core. The board shows whatever the fingerprint reports, so if your M5 is a 10-core-GPU config this may be an undercount in `_machine.py`'s GPU-core detection (the perf numbers are unaffected -- it's just the metadata label). No need to change anything here; just flagging it in case it's worth a follow-up.

Merging once CI is green. Thanks for contributing!

danlee2002:
> Welcome, @danlee2002, and thanks -- this is the first **base M5** on the board (decode 228 tok/s, a nice cross-check against the M5 Pro). 🎉
> 
> I pushed a small merge to your branch: another datapoint (M4) landed just before yours, so the regenerated `ROOFLINES.md` / `rooflines.json` conflicted. I merged `main` in and re-ran `python3 bench/aggregate_rooflines.py` so the tables now include every machine -- nothing changed in your actual measurement, just the generated aggregates.
> 
> One thing to double-check when you get a chance: the fingerprint auto-detected **8 GPU cores** (`gpu_cores: 8`), but the PR title says 10-core. The board shows whatever the fingerprint reports, so if your M5 is a 10-core-GPU config this may be an undercount in `_machine.py`'s GPU-core detection (the perf numbers are unaffected -- it's just the metadata label). No need to change anything here; just flagging it in case it's worth a follow-up.
> 
> Merging once CI is green. Thanks for contributing!

I just took a look at the system settings of my Mac and can confirm that was a mistake on my end. I didn't read my system settings carefully and accidentally mixed up GPU with CPU cores. Sorry about that and let me know if there is anything I need to update.

Issue #127, opened 2026-07-31
tests: run the module self-test batteries under pytest


sbryngelson:
Several modules ship a `_selftest` / `main()` battery of supported-vs-rejected cases (`aneforge/einsum.py`, and the `main()` demos in `linalg.py` / `dsp.py`) that nothing in CI runs. That gap is not hypothetical: a stale `ellipsis` entry sat in einsum's `rejected` list, wrong since `_expand_ellipsis` landed, and went unnoticed precisely because `python aneforge/einsum.py` is never executed in CI (surfaced in #119).

Wire these batteries into the pytest suite so a stale entry fails CI instead of lurking: a thin `tests/test_*_selftest.py` that imports the battery's case lists and asserts the supported ones pass and the rejected ones raise, with the dispatching parts gated behind `requires_ane` (the shape/rejection assertions stay off-device). Where a battery only duplicates what `tests/` already covers, the cleaner move is to delete it and keep the canonical tests -- call that out in the PR rather than wrapping dead weight.

The off-device assertions run in CI on any machine; the dispatching ones need an Apple Silicon Mac, so gate them and note in the PR that they passed locally.

PR #200, opened 2026-08-11
test: cover einsum self-test cases

kocaemre:
## Summary

- move the einsum self-test case inventory to module-level constants
- add pytest coverage that imports those cases and checks supported equations lower off-device with numpy-compatible shapes
- add pytest coverage that the rejected self-test equations still raise before dispatch

## Related issue

Related to #127. This is the off-device slice for the `aneforge/einsum.py` self-test battery; it does not claim on-device numeric coverage.

## Verification

- `pytest -m 'not requires_ane' tests/test_einsum_selftest.py -q`
- `ruff check`
- `python -m pylint --disable=all -e W0311 -e C0303 --indent-string='  ' --recursive=y aneforge tests`
- `pyright aneforge` after installing `.[dev,models] scipy`
- `python -m compileall -q aneforge/einsum.py tests/test_einsum_selftest.py`
- `pytest -m 'not requires_ane' -q`

On-device ANE numeric self-test was not run in this Linux environment.

### DIFF
--- a/aneforge/einsum.py
+++ b/aneforge/einsum.py
@@ -244,6 +244,50 @@ def _expand_to(t: Tensor, present: list, target: list, _sizes) -> Tensor:
   return t.reshape(*shape)
 
 
+# self-test case inventory
+
+_SELFTEST_SUPPORTED = [
+  ("matmul",            "ij,jk->ik",        [(8, 16), (16, 12)]),
+  ("matmul-implied",    "ij,jk",            [(8, 16), (16, 12)]),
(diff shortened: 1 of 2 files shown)


sbryngelson:
Thanks, @kocaemre -- nice first contribution. Hoisting the battery to `_SELFTEST_SUPPORTED` / `_SELFTEST_REJECTED` and parametrizing over it means the einsum equation coverage now runs in CI (off-device, so it works without an ANE) instead of only in the manual `_selftest()`. Behavior of `_selftest()` is unchanged. Merging.

PR #205, opened 2026-08-13
feat(llm): add a Gemma adapter - GeGLU MLP, scaled embeddings, (1+w) RMSNorm

### LINKED ISSUE #193: llm: Gemma family adapter for from_pretrained
Follow-on to the Mistral adapter: add a **Gemma** family adapter to `aneforge.llm`'s `ADAPTERS` registry (`aneforge/llm.py`) so `google/gemma-2b` / `gemma-7b` load via `llm.from_pretrained`.

**Gemma specifics to handle** (vs the Llama `_dense_adapter`):
- Embeddings are **scaled by sqrt(hidden_size)** at input.
- **GeGLU** FFN (gelu, not silu) -- the activation differs from Llama's SwiGLU.
- **Tied embeddings** (the lm_head reuses the input embedding matrix).
- RMSNorm uses a `(1 + weight)` convention in some releases -- check against the HF forward.

**Approach:** `def _gemma_adapter(cfg, sd)` maps the config/weights into `LlamaConfig` + weights, applying the embedding scale and the activation/tying differences, and register `(lambda c: c.model_type in ("gemma", "gemma2"), _gemma_adapter)`.

**Oracle:** `transformers.AutoModelForCausalLM` -- prefill argmax + a short greedy generation must agree. Verify on any Apple Silicon Mac and note it in the PR.

axiom-of-choice:
Closes #193

Gemma is Llama-shaped except for three weight-level differences, all baked into the adapter:

- **Embeddings scaled by sqrt(hidden_size)** at input (HF bakes this into `GemmaTextScaledWordEmbedding`).
- **GeGLU FFN** (`gelu_pytorch_tanh`), which the runner didn't have: new `_geglu` builder composed from fused ops (`muls`/`adds`/`scaled_tanh`), registered in `PREFILL_MLPS`/`DECODE_MLPS` and selected via `LayerSpec(mlp="geglu")`.
- **RMSNorm `(1 + weight)` convention** (checked against `modeling_gemma.py`: `output * (1.0 + weight)`), applied at adapter time so the runner's direct-gamma norm stays untouched.

GQA and the explicit `head_dim` (gemma-7b: 256 != dim // heads) were already handled by the runner. Registered `model_type == "gemma"` only: gemma2 additionally needs logit softcapping, which the runner does not support yet, so registering it would silently produce wrong logits. That is a follow-up.

### Validation

- **On-device (M2 Pro, macOS 26.5.2)**: `google/gemma-2b` int8. Prefill argmax matches HF fp16 (`"The capital of France is"` -> 476 `"a"`); 12-token greedy decode is token-identical to `transformers` fp16: *"a city of contrasts. It is a city of history,"*.
- **In-process test** (`test_prefill_matches_huggingface_gemma`): small random `GemmaForCausalLM` (GQA 4/1, head_dim 16, tied embeddings), cosine > 0.99 + argmax match vs HF. Runs on any ANE; `requires_ane` marked.
- `tests/test_llm.py`: 12 passed. Off-device suite: 303 passed, 2 skipped. ruff, pylint 2-space gate, pyright clean (12 pre-existing errors in `onnx.py`/`_gguf.py` untouched).
- Docs: `docs/llm.md` scope section lists Gemma.

### DIFF
--- a/aneforge/llm.py
+++ b/aneforge/llm.py
@@ -150,6 +150,20 @@ def _swiglu(h: Tensor, w: dict, cfg: LlamaConfig, ls: LayerSpec) -> Tensor:
   return h + (hn.linear(w["wgate"]).silu() * hn.linear(w["wup"])).linear(w["wdown"])
 
 
+def _gelu_tanh(x: Tensor) -> Tensor:
+  """PyTorch 'tanh' GELU (Gemma's `gelu_pytorch_tanh`): 0.5*x*(1+tanh(sqrt(2/pi)*(x+0.044715*x^3)))."""
+  x3 = (x * x) * x
+  inner = (x3 * 0.044715) + x
+  t = inner.scaled_tanh(1.0, 0.7978845608028654)
+  return (x * t.adds(1.0)) * 0.5
+
+
+def _geglu(h: Tensor, w: dict, cfg: LlamaConfig, ls: LayerSpec) -> Tensor:
+  """GeGLU MLP with its residual: `h + down(gelu_tanh(gate(norm(h))) * up(norm(h)))`."""
+  hn = h.rms_norm(w["mlp_norm"], cfg.norm_eps)
+  return h + (_gelu_tanh(hn.linear(w["wgate"])) * hn.linear(w["wup"])).linear(w["wdown"])
+
+
 def prefill_block(x: Tensor, w: dict, cfg: LlamaConfig, cos, sin, ls: LayerSpec | None = None) -> Tensor:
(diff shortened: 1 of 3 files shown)


sbryngelson:
Tested on an M5 Pro: `tests/test_llm.py` is 12/12 on the ANE, including the new `test_prefill_matches_huggingface_gemma` (small random `GemmaForCausalLM`, cosine > 0.99 + argmax match). The scaled-embed / `(1+w)` RMSNorm / GeGLU path all validate against HF. Merging.

One minor follow-up (not blocking): this adds `_rope_scaling(c)` but nothing calls it -- it's dead (ruff won't flag an unused module-level function). Since #206 also touches the RoPE helpers, easiest to consolidate/remove it when that rebases.

Issue #149, opened 2026-08-01
layer_norm: ANE compile/execute failures are affine-value-dependent, not a shape frontier


sbryngelson:
Split out from #146 (which was the `roofline_analysis` harness bug, now fixed by #147). This is the underlying *capability* observation that triggered it.

On an **Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2**, the `layer_norm` bandwidth archetype fails to compile on the ANE at `[R, 4096]` shapes during `bench/device_bandwidth_roofline.py`:

```
compile(mask=0x4) err=11
RuntimeError: ane_e5rt_program_compile failed (...)
E5RT: MILCompilerForANE error: failed to compile ANE model using ANEF.
      Error=_ANECompiler : ANECCompile() FAILED.
```

Two caveats from @axiom-of-choice's investigation, both important:

- **Intermittent.** `[2048, 4096]` compiled on one attempt and failed on a later one, so this is not a clean per-shape threshold.
- **Confounded by the compile breaker.** After a failure the breaker pacing warning fires (`pacing this compile by 15.0s - a compile failed recently`), so a first failure appears to drag subsequent compiles down -- meaning naive per-shape sweeps give unreliable conclusions. Any investigation needs to control for the breaker (fresh process per shape, or reset the breaker between attempts).

Not yet reproduced on other generations. Whether this is a genuine `layer_norm`-on-ANE shape limit, a transient compiler state, or a memory/pacing artifact is open. A careful, breaker-controlled sweep (one shape per fresh process) across a couple of chips would settle it.

Credit: reported by @axiom-of-choice in #146.

axiom-of-choice:
Ran the breaker-controlled sweep you asked for: one compile per fresh subprocess with `ANEFORGE_DISABLE_COMPILE_BREAKER=1`. **Both caveats in the issue turn out to be wrong, and they were my caveats, so this corrects my own report in #146.**

Apple M2 Pro (Mac14,12 Mac mini), macOS 26.5.2, `d6de531`.

**1. It is not intermittent.** Every shape is deterministic across repeats:

```
[256,3993] x5: OK OK OK OK OK
[256,3994] x5: FAIL FAIL FAIL FAIL FAIL
[256,3995] x5: OK OK OK OK OK
[256,3996] x5: FAIL FAIL FAIL FAIL FAIL
```

[...]

PR #206, opened 2026-08-13
feat(llm): add a Mistral adapter (GQA + sliding window) and fix RoPE base under transformers 5.x

### LINKED ISSUE #192: llm: Mistral family adapter for from_pretrained
`aneforge.llm.from_pretrained(name)` loads a causal LM for ANE decode via an **adapter registry** (`ADAPTERS: list` in `aneforge/llm.py`, currently just the `_dense_adapter` Llama default). Each family is a `(predicate, adapter_fn)` pair mapping the HF config + state_dict to `LlamaConfig` + the weight dict. This issue adds a **Mistral** adapter so `mistralai/Mistral-7B-*` and friends load.

**Approach:** Mistral is Llama-shaped (RMSNorm pre-norm, SwiGLU FFN, RoPE) -- the `_dense_adapter` is close. The specifics to handle: grouped-query attention (`num_key_value_heads < num_attention_heads`, so K/V projections are narrower) and the sliding-window attention config (for decode with a resident KV cache the window can be treated as the cache length). Add `def _mistral_adapter(cfg, sd)` returning `(LlamaConfig(...), weights)` and register `(lambda c: c.model_type == "mistral", _mistral_adapter)` in `ADAPTERS`.

**Oracle:** `transformers.AutoModelForCausalLM`. Check the first-token logits (prefill argmax) match, then that a short greedy generation agrees. Verify on any Apple Silicon Mac; a ~7B model needs enough unified memory (int8 helps). Note the local result in the PR.

axiom-of-choice:
Closes #192 (draft - more validation runs pending)

Mistral is Llama-shaped with the same state_dict names, so the adapter reuses the dense plan with grouped-query attention (already handled by the runner) and surfaces the sliding window in `cfg.extra`: with a resident KV cache the window is the cache length, so decode stays exact while `max_len` does not exceed `sliding_window`.

Two fixes along the way, required for 7B models:

- **transformers 5.x RoPE base**: `rope_theta` moved into `config.rope_parameters`; the old `getattr(c, "rope_theta", ...)` silently returned the 10000.0 fallback for every Llama/Qwen/Mistral config, so Qwen3 was being loaded with base 1e4 instead of 1e6. New `_rope_theta()` helper reads the dict with a legacy fallback (also applied to `_dense_adapter` and `_cfg_from_hf`).
- **fp16 state dict in `from_pretrained`**: extracting fp32 cost ~28 GB on top of the loaded model for a 7B - enough to OOM a 32 GB Mac. The weights are baked to fp16 (or quantized) for the ANE anyway, so extract fp16 directly.

### Validation so far (M2 Pro, macOS 26.5.2)

- Small in-process `MistralForCausalLM` test (GQA 4/2, sliding_window 32): cosine > 0.99 + argmax match vs HF. Green and reproducible.
- Real `mistralai/Mistral-7B-Instruct-v0.3` int8: **prefill argmax matches HF exactly** ("The capital of France is" -> 6233 "Paris").
- The decode-program compile is **flaky on this machine** for 7B-sized graphs (`ANECCompile() FAILED`, err=11 - the same instability class as #149). It compiled once (greedy agreed for the first 4 tokens before a near-tie drift) and failed ~6 consecutive retries after that.

**Draft**: running the decode validation on a fresh machine to close the greedy check and quantify the near-tie drift. The adapter mapping itself is verified by the prefill argmax match and the in-process test.

### DIFF
--- a/aneforge/llm.py
+++ b/aneforge/llm.py
@@ -494,6 +494,22 @@ def _gemma_adapter(c, sd) -> tuple[LlamaConfig, dict]:
   return cfg, w
 
 
+def _mistral_adapter(c, sd) -> tuple[LlamaConfig, dict]:
+  """Adapter for Mistral decoders: Llama-shaped (same state_dict names) with grouped-query attention
+  and a sliding window. With a resident KV cache the window is the cache length, so decode stays
+  exact while `max_len` does not exceed `sliding_window`; the limit is surfaced in `cfg.extra`."""
+  n = int(c.num_hidden_layers)
+  cfg = LlamaConfig(dim=c.hidden_size, n_layers=n, n_heads=c.num_attention_heads,
+                    n_kv_heads=int(getattr(c, "num_key_value_heads", c.num_attention_heads)),
+                    ffn_dim=c.intermediate_size, vocab=c.vocab_size,
+                    rope_base=_rope_theta(c), norm_eps=float(c.rms_norm_eps),
+                    head_dim=int(getattr(c, "head_dim", 0) or 0),
+                    layers=[LayerSpec(mixer="attention", mlp="swiglu") for _ in range(n)])
+  if getattr(c, "sliding_window", None):
+    cfg.extra["sliding_window"] = int(c.sliding_window)
+  return cfg, _weights_from_state_dict(sd, cfg)
+
+
 def _cfg_from_hf(c) -> LlamaConfig:
   """Back-compat: a `LlamaConfig` (no per-layer plan) from a Hugging Face config."""
   return LlamaConfig(dim=c.hidden_size, n_layers=c.num_hidden_layers, n_heads=c.num_attention_heads,
@@ -507,6 +523,7 @@ def _cfg_from_hf(c) -> LlamaConfig:
 # wins, dense Llama/Qwen is the fallback. New archs append here (e.g. aneforge.moe) so the loader stays generic.
 ADAPTERS: list = [
   (lambda c: c.model_type == "gemma", _gemma_adapter),
+  (lambda c: c.model_type == "mistral", _mistral_adapter),
 ]
 
 
--- a/tests/test_llm.py
+++ b/tests/test_llm.py
@@ -1,6 +1,6 @@
 import numpy as np
 import aneforge as af
-from aneforge.llm import LlamaConfig, LlamaPrefill, prefill_block, rope, rope_tables, _weights_from_state_dict, _cfg_from_hf, _gemma_adapter
+from aneforge.llm import LlamaConfig, LlamaPrefill, prefill_block, rope, rope_tables, _weights_from_state_dict, _cfg_from_hf, _gemma_adapter, _mistral_adapter
 from _helpers import requires_ane, make_random_llama_model as _random_model
 
 pytestmark = requires_ane  # every test in this module compiles/dispatches to the ANE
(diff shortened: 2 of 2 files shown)


axiom-of-choice:
The decode-program compile for the 7B graph is flaky on this machine (ANECCompile() FAILED, err=11, the same instability class as #149): the same MIL compiled successfully once (prefill argmax matched HF and greedy agreed for the first 4 tokens) and then failed ~6 consecutive retries. Running more validation on a fresh machine to close the greedy check and quantify the near-tie drift at token 5.

axiom-of-choice:
**Validation run on a fresh M2 Pro (32 GB, macOS 26.5.2)** — `Mistral-7B-Instruct-v0.3`, int8, retry loop around the decode-program compile. The 7B decode program compiled on the first attempt here (no `ANECCompile() FAILED` on this machine, unlike the earlier run).

**Greedy decode, 12 tokens — side by side:**

| model | output |
|---|---|
| **HF (fp16 oracle)** | `Paris, but the capital of the French language is Quebec City` |
| **ANE (int8, fp16 activations)** | `Paris, but the country's second largest city is Mar` |

The first 4 tokens are identical; the models diverge at token 5 and each continues its own coherent sentence.

**Prefill:** argmax exact (6233 "Paris"), logits cosine **0.98944**.

**Per-step table** — position i = the i-th generated token; both sides condition on the same preceding tokens (which agree through position 4):

| pos | ANE argmax | HF argmax | match | gap (ANE top-2) | cos |
|---|---|---|---|---|---|
| 1 | 6233 'Paris' | 6233 'Paris' | ✅ | 0.0555 | 0.98944 |
| 2 | 29493 ',' | 29493 ',' | ✅ | 0.8801 | 0.99156 |
| 3 | 1330 'but' | 1330 'but' | ✅ | 0.6829 | 0.98855 |
| 4 | 1040 'the' | 1040 'the' | ✅ | 0.5400 | 0.98872 |
| 5 | 3707 'country' | 6333 'capital' | ❌ | 0.3937 | 0.98502 |
| 6 | 29577 '’' | 1070 'of' | ❌ | 0.1342 | 0.93636 |
| 7 | 29481 's' | 1040 'the' | ❌ | 7.6726 | 0.84452 |
| 8 | 2444 'second' | 5717 'French' | ❌ | 0.7261 | 0.92103 |
| 9 | 8407 'largest' | 4610 'language' | ❌ | 0.5090 | 0.91168 |
| 10 | 3758 'city' | 1117 'is' | ❌ | 3.2113 | 0.92563 |
| 11 | 1117 'is' | 28566 'Quebec' | ❌ | 0.8925 | 0.91664 |
| 12 | 2239 'Mar' | 4573 'City' | ❌ | 1.9344 | 0.90221 |

**The drift at pos 5 is a genuine near-tie, not a systematic error.** Both models rank exactly the same top-2 — "country" (3707) and "capital" (6333): ANE orders them `country > capital` by **0.3937** logits; HF orders them `capital > country` by **0.5312**. The two candidates sit ~0.92 logits apart across both models, so the int8 + fp16-ANE path flips the ordering. After the flip, cosine decays (0.985 → 0.84–0.93) simply because each model conditions on its own different continuation from that point on.

**Deterministic**: identical token sequence across two independent runs (compile cache reused).

axiom-of-choice:
Marked ready for review. Also measured throughput on the same M2 Pro while the graphs were warm (compile cache reused within the process):

| metric | value |
|---|---|
| decode, 7B int8, warm | **~1.4 tok/s** (16 tokens in 11.2 s) |
| prefill, 6-token prompt, warm | ~0.8 s |
| one-time per fresh process (int8 bake + Espresso compile) | ~2 min before the first token |

**The number that frames it:** the M2 Pro row in the decode roofline (276 tok/s) is measured on the ~0.13B bench model. 7B decode is weight-streaming-bound — every token pushes ~7.5 GB of int8 weights through the ANE's memory path, the known weak point of this family (0.98 GB/s streaming on the M2 Pro roofline vs 7.4 on M1 Max, 20 on M5). So ~1.4 tok/s is the hardware regime for 7B on this chip, not an adapter regression. On an M5 the same bound scales to roughly ~3 tok/s by bandwidth ratio.

One new observation: past token ~13 the generation enters a **deterministic repetition loop** — it re-emits the prompt ("…second largest city is **Marse The capital of France is Paris, but the country's second largest city is**…"), identical in both the 16- and 32-token runs. Not seen in the 12-token validation (which stops at the drift point). If longer generations are a goal, worth its own look.

sbryngelson:
Tested on an M5 Pro. The adapter mapping is solid: `tests/test_llm.py` is 12/12 on the ANE including `test_prefill_matches_huggingface_mistral` (GQA 4/2, sliding_window surfaced in `cfg.extra`, cosine > 0.99 + argmax match). Nothing here is in doubt except the 7B decode compile you flagged, which is the #149-class `ANECCompile()` err=11 instability, not this PR's code.

**The RoPE base fix is more important than the Mistral adapter, and it's a live bug.** I confirmed it independently on transformers 5.9.0: a modern config has no `.rope_theta` attribute at all (it's in `rope_parameters`), so the old `getattr(c, "rope_theta", 10000.0)` silently returns **10000.0** for every model with a non-default base. Concretely, old code -> `10000.0`, `_rope_theta()` -> `1000000.0`. That means **every Qwen3 (base 1e6) and Llama-3 (base 5e5) load on transformers 5.x has been getting the wrong RoPE base** through `_dense_adapter` / `_cfg_from_hf` -- silent wrong logits, unrelated to Mistral.

Suggestion: **split the two infra fixes (`_rope_theta` + the fp16 state-dict OOM fix) into their own small PR** so they land now -- they're verified and fix existing models -- and keep this one as the Mistral adapter draft until the 7B decode re-validation closes. That also sidesteps the conflict: #205 just merged and also added `_rope_theta` + flipped `ADAPTERS`, so this branch needs a rebase regardless.

Happy to fast-track the split-out RoPE/fp16 PR if you'd rather I open it and credit your diagnosis. Great find either way.

sbryngelson:
Opened #207 with the two infra fixes split out (RoPE base + fp16 state dict), credited to you -- set to auto-merge on green. Once it lands, this branch can rebase onto main and drop those two hunks, leaving just the Mistral adapter + the sliding-window `cfg.extra` for the 7B decode re-validation. Thanks again for finding the RoPE base issue; it was silently wrong for every Qwen3/Llama-3 load on transformers 5.x.

axiom-of-choice:
> Opened #207 with the two infra fixes split out (RoPE base + fp16 state dict), credited to you -- set to auto-merge on green. Once it lands, this branch can rebase onto main and drop those two hunks, leaving just the Mistral adapter + the sliding-window `cfg.extra` for the 7B decode re-validation. Thanks again for finding the RoPE base issue; it was silently wrong for every Qwen3/Llama-3 load on transformers 5.x.

Rebased!

sbryngelson:
Re-tested after your rebase (M5 Pro): `test_llm.py` 13/13 on the ANE including `test_prefill_matches_huggingface_mistral` (GQA + sliding_window surfaced in `cfg.extra`, cosine > 0.99 + argmax). The rebase is clean -- RoPE/fp16 are gone (landed in #207). The 7B decode compile flakiness you flagged is the #149-class ANECCompile instability, independent of this adapter, which validates on prefill argmax + the in-process test. Merging.

PR #208, opened 2026-08-14
models: support RoBERTa/XLM-R rerankers in CrossEncoder

### LINKED ISSUE #190: models: bge-reranker (RoBERTa/XLM-R) support in CrossEncoder
`CrossEncoder` (`aneforge/models.py`, added in #189) reranks (query, passage) pairs on the ANE for **BERT-family** sequence-classification models (pooler + classifier head). It raises a clear error on RoBERTa/XLM-R models. This issue adds support for those, so the most popular reranker -- [`BAAI/bge-reranker-base`](https://huggingface.co/BAAI/bge-reranker-base) / `-large` (XLM-RoBERTa) -- runs on the engine.

**What differs from BERT** (all host-side plumbing; the transformer graph is unchanged):
- **No token-type embeddings.** RoBERTa uses a single segment; skip the `typ` lookup.
- **Position ids start at `padding_idx + 1`** (offset), not 0.
- **Head is `RobertaClassificationHead`**, not pooler+classifier: `classifier.dense` (D->D) + tanh + `classifier.out_proj` (D->num_labels), applied to the first token, not a separate pooler.

**Approach:** detect the RoBERTa family (e.g. `{pref}.embeddings.token_type_embeddings` absent or `classifier.out_proj.weight` present), branch the embedding + head accordingly, and reuse the existing encoder `_build`. The `_BERT_KEYS` layer names are shared between BERT and RoBERTa, so the transformer loop is the same.

**Oracle:** `transformers.AutoModelForSequenceClassification`. Compare scores + ranking on a few (query, passage) sets. Verify on any Apple Silicon Mac and note the local result in the PR (CI has no ANE).

axiom-of-choice:
Closes #190.

`CrossEncoder` handled BERT-family sequence-classification models and raised on RoBERTa/XLM-R, which left out the most widely used reranker family. `BAAI/bge-reranker-base` now runs on the engine.

The encoder graph and the `_BERT_KEYS` layer names are already shared between the two families, so the whole difference is host-side plumbing and it lands as weight selection rather than a second code path. `predict()` and `_build()` are unchanged.

## What changed

**`_seqcls_head(sd, pref)`** picks the classification head by state-dict shape. BERT is `pooler.dense -> tanh -> classifier`. RoBERTa is `RobertaClassificationHead`, i.e. `classifier.dense -> tanh -> classifier.out_proj`, with no pooler at all. Both are the same two-layer tanh head on token 0, so once the four keys are chosen the arithmetic is identical.

**`_position_ids(ids, pad_id)`** implements HF `create_position_ids_from_input_ids`: `cumsum(ids != pad) * mask + pad`. RoBERTa counts from `padding_idx + 1` and skips pads, which is why its position table carries `max_position_embeddings + 2` rows (514 here). Counting from 0 reads the wrong row for every token.

The token-type lookup now falls back to zeros for a checkpoint that genuinely omits the table.

Both helpers are module-level so CI can cover them off-device.

## One correction to the issue

The issue suggests detecting the family by `{pref}.embeddings.token_type_embeddings` being absent. That does not work: `bge-reranker-base` **does** ship `token_type_embeddings`, shape `(1, 768)`, a single row that is always index 0. Absence identifies nothing, and keying on it also misclassifies BERT. Detecting by head shape covers both families, and `test_roberta_is_not_detected_by_a_missing_token_type_table` pins that down so the shortcut is not reintroduced.

Everything else was read off the model rather than assumed: `model_type=xlm-roberta`, `num_labels=1`, `pad_token_id=1`, `type_vocab_size=1`, no `roberta.pooler.*`, and the tokenizer returns no `token_type_ids` (the existing fallback already covers that).

## Verification

**On device (M2 Pro, macOS 26.5.2), `BAAI/bge-reranker-base` against `AutoModelForSequenceClassification`:**

| query | passage | HF | ANE | abs err |
|---|---|---|---|---|
| what is a panda? | The giant panda is a bear species endemic to China. | 0.9140 | 0.9113 | 0.00273 |
| what is a panda? | Paris is the capital and most populous city of France. | -10.1960 | -10.1959 | 0.00008 |
| how do I boil an egg? | Place eggs in a pot, cover with water, bring to a boil. | 6.3916 | 6.3929 | 0.00131 |
| how do I boil an egg? | The Apple Neural Engine accelerates matrix multiplication. | -10.1930 | -10.1928 | 0.00021 |

Max abs err 0.00273 on raw logits, ranking identical, relevant above irrelevant on both sides.

**BERT path unchanged** (`cross-encoder/ms-marco-MiniLM-L-6-v2`, the path added in #189): max abs err 0.00236, same order, and `pad_id` is `None` so it takes the BERT branch.

**Mutation, behavioural rather than by import:**

- `_position_ids` ignoring `pad_id`: the 2 position tests fail.
- detecting RoBERTa by a missing token-type table: 4 tests fail, including the BERT one.
- restored: 10/10.

`pytest -m "not requires_ane"`: 321 passed, 2 skipped. ruff, pylint 2-space 10.00/10 and compileall clean; pyright reports the same 12 baseline errors in `onnx.py`/`_gguf.py`.

## Risks and scope

- The end-to-end check needs an ANE and a model download, so it cannot live in CI. What CI does cover is the head selection and the position indexing, which is where a regression would land.
- DistilBERT-style heads are still rejected, with the error message updated to say what is supported.
- `#189` added `CrossEncoder` and `ViT` with no tests. This PR opens `tests/test_cross_encoder.py`; `ViT` is still uncovered, happy to follow up if useful.

### DIFF
--- a/aneforge/models.py
+++ b/aneforge/models.py
@@ -153,11 +153,48 @@ class Encoder:
     return np.asarray(vecs, dtype=np.float32)
 
 
+def _position_ids(ids: np.ndarray, pad_id: int | None) -> np.ndarray:
+  """Position ids as the reference implementations build them.
+
+  BERT counts from 0. RoBERTa counts from `padding_idx + 1` and skips pads, i.e. HF's
+  `create_position_ids_from_input_ids`: `cumsum(ids != pad) * mask + pad`. That offset is why a
+  RoBERTa position table has `max_position_embeddings + 2` rows; counting from 0 reads the wrong row
+  for every token and silently shifts the whole sequence."""
+  if pad_id is None:
+    return np.arange(len(ids))
+  mask = (ids != pad_id).astype(np.int64)
+  return np.cumsum(mask) * mask + pad_id
+
+
+def _seqcls_head(sd: dict, pref: str | None) -> tuple[str, str, str, str] | None:
+  """The two-layer tanh classification head, as (dense_w, dense_b, out_w, out_b) state-dict keys.
+
+  BERT is `pooler.dense -> tanh -> classifier`; RoBERTa is `RobertaClassificationHead`, i.e.
+  `classifier.dense -> tanh -> classifier.out_proj`, with no pooler at all. Same arithmetic on token 0,
+  different names. Returns None when neither shape is present.
+
+  RoBERTa is identified by its head, *not* by a missing token-type table: XLM-R does ship
+  `token_type_embeddings` (a single row, always index 0), so absence identifies nothing."""
+  if "classifier.dense.weight" in sd and "classifier.out_proj.weight" in sd:
+    return ("classifier.dense.weight", "classifier.dense.bias",
+            "classifier.out_proj.weight", "classifier.out_proj.bias")
+  if pref is not None and f"{pref}.pooler.dense.weight" in sd and "classifier.weight" in sd:
+    return (f"{pref}.pooler.dense.weight", f"{pref}.pooler.dense.bias",
+            "classifier.weight", "classifier.bias")
+  return None
+
+
 class CrossEncoder:
-  """Reranker: scores (query, passage) pairs with a BERT-family sequence-classification
+  """Reranker: scores (query, passage) pairs with a BERT- or RoBERTa-family sequence-classification
   model, the transformer running on the ANE. Mirrors `sentence_transformers.CrossEncoder`:
   `CrossEncoder(name).predict([(query, passage), ...])` returns one relevance score per pair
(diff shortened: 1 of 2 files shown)


sbryngelson:
Tested on M5 Pro: the 10 off-device helper tests pass (head-detection + RoBERTa position-id logic, including the correction that `token_type_embeddings` presence can't distinguish the families). Confirmed no regression to the shared path -- the cached BERT reranker still runs on the ANE at relerr 2e-4 with ranking preserved. Good call detecting by head shape rather than the token-type table. Merging.

PR #209, opened 2026-08-14
models: generalize the ResNet loader to 18/34/50/101

### LINKED ISSUE #191: models: ResNet image-classifier loader (resnet18/34/50)
`Vision` (`aneforge/models.py`) loads torchvision **ResNet-18** as one fused ANE program, folding BatchNorm into the preceding conv (`_fold`). This issue generalizes it to a `load_resnet(name_or_depth)` covering the popular ResNet depths.

**Scope:** resnet18 / resnet34 (BasicBlock, already the `_block` shape) and resnet50 / resnet101 (Bottleneck: 1x1 -> 3x3 -> 1x1 with a 4x channel expansion). torchvision weights (`IMAGENET1K_V1/V2`) or timm.

**Approach:** the BasicBlock path already exists in `Vision._block`; add a `_bottleneck` block and parameterize the stage layout by depth (`{18:(2,2,2,2), 34:(3,4,6,3), 50:(3,4,6,3)}` with block type per depth). Reuse `_fold` for every conv+BN pair. The stem, global average pool, and `fc` head are shared.

**Oracle:** `torchvision.models.resnet{18,34,50}(weights=...)`. Compare logits / top-1 on a fixed image (fp16 vs fp32; expect small divergence). Verify on any Apple Silicon Mac and note it in the PR.

axiom-of-choice:
Closes #191.

`Vision` hardcoded ResNet-18: two BasicBlocks per stage and a 512-wide head. `af.load_resnet(name_or_depth)` now covers 18/34/50/101 and takes `50`, `"50"` or `"resnet50"`. `af.load_resnet18()` stays as the shorthand and returns bit-identical output.

## What changed

**`_bottleneck`** adds the 1x1 -> 3x3 -> 1x1 block with its 4x expansion. The stride sits on the 3x3, which is torchvision's ResNet V1.5 variant rather than the placement in the original paper.

**`_shortcut`** decides the residual branch from the weights rather than from a stage rule. torchvision projects when `stride != 1 or inplanes != planes * expansion`.

**The head width** comes from `fc.weight` instead of a constant 512.

`_RESNETS` holds the stage layouts, and `_resnet_depth` does the parsing so an unsupported depth fails with a list of what is supported instead of a `KeyError` deep in the build.

## Two things worth flagging

**Stage 1 is projected for Bottleneck and not for BasicBlock.** The existing rule was `downsample=(i == 0 and name != "layer1")`, which is correct for ResNet-18 and wrong for 50/101: there `layer1.0` changes 64 channels into 256 at stride 1, so it does carry a projection. Reading `downsample.0.weight` from the state dict makes the rule unnecessary and covers both families. `test_bottleneck_projects_stage_one_and_basicblock_does_not` pins the underlying fact against torchvision.

**Top-1 agreement is too weak a check for this port.** Moving the Bottleneck stride from the 3x3 to the first 1x1, i.e. building V1 instead of V1.5, leaves every tensor shape intact and *still gives the same top-1* on my test image, while cosine against torchvision drops from 0.999999 to 0.9499. That is why the numbers below are cosine and full-logit deltas rather than a label match, and why the V1.5 placement is called out in a comment and in `docs/developer/models.md`.

## Verification

**On device (M2 Pro, macOS 26.5.2), against `torchvision.models.resnet*` with `IMAGENET1K_V1`, one fixed input for both sides:**

| depth | block | top-1 | top-5 | max abs logit delta | cosine |
|---|---|---|---|---|---|
| 18 | BasicBlock | match | match | 0.01356 | 0.999998 |
| 34 | BasicBlock | match | match | 0.01810 | 0.999998 |
| 50 | Bottleneck | match | match | 0.02213 | 0.999999 |
| 101 | Bottleneck | match | match | 0.03028 | 0.999994 |

**Back-compat:** `load_resnet18()` and `load_resnet(18)` return bit-identical logits (`np.array_equal`).

**Mutation:**

- shortcut keyed on the stage index instead of the weights: `test_shortcut_reads_the_weights_rather_than_a_stage_rule` fails.
- Bottleneck stride moved to conv1: no test fails, cosine on ResNet-50 drops to 0.9499 while top-1 still matches. Reported above rather than hidden.
- restored: 16/16.

`pytest -m "not requires_ane"`: 327 passed, 2 skipped (up from 311). ruff, pylint 2-space 10.00/10 and compileall clean; pyright reports the same 12 baseline errors in `onnx.py`/`_gguf.py`.

## Notes

- `[dev]` does not install torchvision, it lives in the `models` extra, so the tests that need it use `importorskip` and skip in CI rather than fail. The layout table itself is pinned by a test that runs everywhere, and a second test checks that table against the torchvision factories when it is present. Those torchvision tests construct with `weights=None`, so they need no download.
- Docs updated: `docs/aneforge-api.md`, `docs/developer/models.md`, `docs/getting-started.md`, and the `models` extra comment in `pyproject.toml`.
- ResNeXt and wide_resnet reuse Bottleneck with `groups`/`base_width`, so they are close but not free; left out since the issue scopes to plain ResNet. Happy to add if wanted.

### DIFF
--- a/aneforge/__init__.py
+++ b/aneforge/__init__.py
@@ -31,7 +31,7 @@ from ._rewrite import reduce_sum_to_matmul, paired_subtract
 from .autograd import (Adam, adam_step, backward, backward_from, conv2d, conv_param,
                        mse, parameter, SGD, softmax_cross_entropy, Trainer, UnrolledTrainer)
 from .streaming import CheckpointedStack
-from .models import Encoder, Vision, load, load_resnet18, conv_block, cifar_cnn, group_norm_train
+from .models import Encoder, Vision, load, load_resnet, load_resnet18, conv_block, cifar_cnn, group_norm_train
 from .onnx import load_onnx, onnx_to_tensor, onnx_to_features
 from .llm import LlamaConfig, LlamaPrefill, from_pretrained as load_llm, rope, rope_tables, prefill_block
 from . import moe as moe   # registers the "moe" MLP + Qwen3-MoE adapter into the llm registries
@@ -52,7 +52,7 @@ __all__ = [
     "precision_risk", "project_peak",
     "CompileBackoffError", "reset_compile_breaker",
     "reduce_sum_to_matmul", "paired_subtract",
-    "load", "load_resnet18", "Encoder", "Vision", "conv_block", "cifar_cnn", "group_norm_train",
+    "load", "load_resnet", "load_resnet18", "Encoder", "Vision", "conv_block", "cifar_cnn", "group_norm_train",
     "Paired", "paired",
     "parameter", "backward", "backward_from", "mse", "SGD", "Adam",
     "softmax_cross_entropy", "Trainer", "UnrolledTrainer", "adam_step",
--- a/aneforge/models.py
+++ b/aneforge/models.py
@@ -23,17 +23,48 @@ def load(name: str, int8: bool = False, pooling: str = "mean") -> "Encoder":
   return Encoder(name, int8=int8, pooling=pooling)
 
 
+# ResNet stage layouts, straight from the torchvision factories: blocks per stage and the residual
+# block shape. Bottleneck expands its output 4x, which is what makes its stage-1 shortcut projected.
+_RESNETS: dict[int, tuple[str, tuple[int, int, int, int]]] = {
+  18: ("basic", (2, 2, 2, 2)),
+  34: ("basic", (3, 4, 6, 3)),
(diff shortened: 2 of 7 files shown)


sbryngelson:
Tested on M5 Pro: 16/16 off-device tests pass (incl. the torchvision layout + stage-1-projection checks), and ResNet-50 -- the new Bottleneck path -- matches torchvision at cosine 0.999999, top-1 agrees. That 6-nines figure also confirms the V1.5 stride placement is right (your note says the V1 misplacement drops it to ~0.95). Merging.

Issue #203, opened 2026-08-12
Bring up GPT-2 on the ANE: the first LayerNorm decoder LLM (unblocked by #201)


sbryngelson:
## Why now

#201 made `layer_norm` compile at `D >= 1024` on the ANE (root cause #162). The models we already run on-device -- Llama / Qwen decoders -- use **RMSNorm**, which was never affected, so that fix does not show up in them. The family it *does* unblock is the classic **GPT-2** decoders: they are pure **LayerNorm** at `D >= 1024`, and were walled by exactly the #162 compile failure until now. So GPT-2 is the natural first "big LLM" to stand up on this new capability.

The pieces already exist -- we have a pre-norm transformer graph (`examples/vit.py`), a tiled vocab head for large `lm_head` (#183), and the Hub-loader pattern in `aneforge/models.py` (BERT, ViT, Llama). This issue is to wire them into a GPT-2 forward + greedy decode and validate against Hugging Face.

## Targets and feasibility

One ANE program bakes its weights in and tops out around ~2 GB (see the program-ceiling notes), so pick the target by fp16 weight size:

| model | params | `D` | layers | fp16 weights | fits one program? |
|---|---|---|---|---|---|
| **gpt2-medium** | 355M | 1024 | 24 | ~0.7 GB | yes -- **primary milestone** |
| gpt2-large | 774M | 1280 | 36 | ~1.5 GB | yes, under the ceiling |
| gpt2-xl | 1.5B | 1600 | 48 | ~3.1 GB | no -- needs per-layer weight streaming (stretch) |

`gpt2-medium` (D=1024) is the clean first target: it directly exercises the newly-unblocked dimension and fits a single fused program. `gpt2-xl` is a stretch that would reuse the streaming/segmented-decode path, not this issue.

## Task

1. Add a `GPT2` loader to `aneforge/models.py` (mirroring the ViT/BERT loaders): load HF `gpt2-medium` weights, map them, and build the graph.
2. Graph: token embed + learned positional embed -> N pre-norm blocks (`ln_1` -> causal MHA -> residual; `ln_2` -> Linear -> GELU -> Linear -> residual) -> final `ln_f` -> `lm_head`. Note GPT-2's vocab is 50257, which exceeds the matmul-dim cap on the A13-A15 families, so the head must use the tiled path from #183 (it fits one tile only on A16/M5 otherwise).
[...]

PR #210, opened 2026-08-14
models: GPT-2 on the ANE (pre-norm decoder, tiled tied lm_head, greedy decode) (#203)

axiom-of-choice:
## What

Bring up **GPT-2** on the ANE (issue #203): the first "large" LLM to run here, unblocked by #201's LayerNorm-at-D>=1024 fix. GPT-2-family models are pure-LayerNorm decoders (Llama/Qwen use RMSNorm), so they are the family #201 actually unlocks.

- `load_gpt2(name, int8=False, max_layers=None)` + `class GPT2` in `aneforge/models.py`, mirroring the `ViT` loader: pre-norm blocks (`ln_1 -> causal SDPA -> res; ln_2 -> Linear -> gelu_new -> Linear -> res`), host-side wte/wpe embedding, `ln_f`, and the **tied lm_head tiled along vocab** (#183): `gpt2-medium`'s 50257 vocab -> 4 output-port tiles `[16384, 16384, 16384, 1105]` on family 3, stitched host-side.
- GPT-2's `Conv1D` weights (`[in, out]`) are transposed to the `.linear()` `[out, in]` convention at load; `gelu_new` (tanh-approx) composed from native ops, mirroring the ONNX handler at `onnx.py:352-360`.
- `examples/gpt2.py`: compiles the full 24-layer model as one fused program (retry + fallback ladder like `examples/vit.py`), validates the forward vs HF fp32, and greedy-decodes K=16 tokens with a **single compiled program**: the growing sequence is padded to `S_MAX` at the end, which is exact under causal attention (positions > j are masked, LayerNorm is row-wise) - verified against HF with `np.allclose`. Compiling once instead of once per length keeps the decode cheap and the disk footprint small.
- `tests/test_gpt2.py` (off-device): Conv1D->linear weight-map orientation, tile shapes, and off-device MIL lowering of the gelu_new composition and the 4-tile head.

## Measured (on-device)

Chip: **M2 Pro / A15 / family 3**, macOS 26.5.2, fp16 activations, weights fp16 (not int8).

`python3 examples/gpt2.py` -> **RESULT: PASS**:

```
PATH: full 24-layer model compiled as one program (S_MAX=22)
pad-equivalence (HF padded row == HF unpadded row): True
VALIDATION: logit cosine (last row) = 1.0000
  ANE top-1 = 287 (' in') | ref top-1 = 287 (' in')
GREEDY DECODE (K=16): 16/16 token-id match vs HF fp32
  ane greedy:  in the hands of the people.\n\nThe future of artificial intelligence is in
  decode: 3.57 s/token (0.28 tok/s, compiles excluded)
```

- `gpt2-medium`'s 50257-vocab head **compiles on A13-A15 (family 3)** via the #183 tiled lm_head (4 tiles of 16384 + 1105); on A16+ (M5) the vocab fits a single 65536-cap tile and the loader takes the single-program path automatically.
- The greedy path was also validated at 1 layer (16/16) and the full-24-layer forward (cos 1.0000, top-1 matching HF) across multiple runs before the final PASS.

## Checks

- CI-gateable: `ruff check`, pylint 2-space gate (10.00), `pyright aneforge` (no new errors; only the pre-existing optional-import class, resolved by the `models` extra the type-check job installs), `compileall`, `pytest -m "not requires_ane"` (247 passed incl. the 4 new tests).
- On-device: `PYTHONPATH=. python3 tests/run_corpus.py` -> GATE GREEN (90/90); the acceptance example above.

## Notes

- One machine-specific finding worth recording: on this volume the ANECompilerService keeps deleted bundle files open until it is restarted, so each 24-layer compile holds ~10 GB of disk until a service restart/reboot. The single-program padded decode was chosen partly for that reason; the loader itself is agnostic.
- The plan's KV-cache per-token decode stays out of scope (recompute-per-length was the original decode choice; the example now runs it through one program instead).


aneforge/models.py
```diff
@@ -319,6 +321,148 @@ def release(self) -> None: self._model.release()
...
+
+def _gelu_new(x: Tensor) -> Tensor:
+  """GPT-2's tanh-approximated GELU, composed from native ops (mirrors the ONNX
+  `approximate="tanh"` handler, aneforge/onnx.py:352-360)."""
+  inner = (x + x.pow(3.0) * 0.044715) * np.sqrt(2.0 / np.pi)
+  return (x * 0.5) * inner.tanh().adds(1.0)
+
+
+# smallest supported matmul output dim on family 3 (A13-A15, M1/M2); see bench/decode_measurement.py:130-135
+_LMHEAD_TILE = 16384
```
axiom-of-choice: Buried

sbryngelson:
Reproduced on an M5 Pro (family differs from your M2 Pro, so the vocab tiling differs) -- `python3 examples/gpt2.py` -> **RESULT: PASS**: forward cosine 1.0000, greedy 16/16 token-id match vs HF fp32, full 24-layer model as one fused program, lm_head tiled 4 ways `[16384, 16384, 16384, 1105]`. Off-device `tests/test_gpt2.py` 7/7. This is exactly what #203 asked for -- the first LayerNorm decoder LLM on the ANE, unblocked by #201.

I rebased onto main for you (it conflicted with #209's `load_resnet` and #212's `load_vit`/`ViT` on the shared `__init__` import + `__all__`) and pushed the union resolution to your branch; README/models.py auto-merged. Merging. Thanks -- great milestone.

PR #221, opened 2026-08-15
models: add decoded GPT-2 generation helper

### LINKED ISSUE #220: models: let GPT2.generate return decoded text, not just token ids
GPT2.generate(prompt, max_new_tokens) returns the generated token ids as a list[int], so a caller has to reach for the tokenizer to see the text. The GPT2 object already holds the tokenizer (self.tok), so decoding is one call.

Task: give GPT2 a way to return the decoded string. Either add a decode-to-text option to generate, or a small generate_text(prompt, max_new_tokens) that calls generate and returns self.tok.decode(ids). Keep the existing token-id return working (a new method, or a keyword that defaults to today's behavior).

Acceptance:
- A caller can get generated text from a GPT2 in one call.
- The existing generate(...) -> list[int] behavior is unchanged (back-compatible).
- examples/gpt2.py or docs/developer/models.md shows the text path.

Pointers: aneforge/models.py (GPT2.generate, GPT2.tok), docs/developer/models.md (load_gpt2 section).

PozziTiv4ik:
## Summary
- add a backward-compatible `GPT2.generate_text()` helper that decodes the token IDs returned by `generate()`
- document both token-ID and decoded-text generation paths
- cover delegation and decoding with an off-device unit test

## Testing
- `pytest -o addopts="" -m "not requires_ane" -q` - 349 passed, 2 skipped, 697 deselected
- `ruff check`
- `python -m pylint --disable=all -e W0311 -e C0303 --indent-string='  ' --recursive=y aneforge tests`
- `pyright aneforge` in a CI/Linux-equivalent type-check environment - 0 errors
- `python -m compileall -q aneforge`
- `mkdocs build --strict`
- package build and `twine check` for both artifacts

Not run: the on-device ANE suite; this change only delegates existing generation output to the host-side tokenizer.

Closes #220

### DIFF
--- a/aneforge/models.py
+++ b/aneforge/models.py
@@ -553,6 +553,10 @@ class GPT2:
       ids = np.concatenate([ids, [int(self(ids)[-1].argmax())]])
     return [int(t) for t in ids[len(base):]]
 
+  def generate_text(self, prompt: str, max_new_tokens: int = 16) -> str:
+    """Greedy-decode and return the newly generated tokens as text."""
+    return self.tok.decode(self.generate(prompt, max_new_tokens))
+
   def release(self) -> None:
     for net, head in self._cache.values():
       net.release(); head.release()
--- a/docs/developer/models.md
+++ b/docs/developer/models.md
@@ -77,7 +77,9 @@ A strided `PxP` patch conv is walled on the ANE, so patch embedding runs as `spa
 
 ```python
 gpt2 = af.load_gpt2("gpt2-medium")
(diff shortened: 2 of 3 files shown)


sbryngelson:
Thanks, and welcome. This is exactly the shape #220 asked for: generate_text() delegates to generate() + the resident tokenizer, the existing list[int] return is untouched, and the Mock-based test covers the delegation off-device so CI runs it without an ANE. Docs show both paths. Merging.

PR #223, opened 2026-08-15
feat: add DistilBERT CrossEncoder support

### LINKED ISSUE #219: models: support DistilBERT rerankers in CrossEncoder
CrossEncoder handles BERT-family and RoBERTa/XLM-R sequence-classification models (#208). Its docstring notes DistilBERT-style heads aren't wired up yet. This adds them, so a DistilBERT reranker (e.g. a cross-encoder/ms-marco DistilBERT checkpoint) runs on the ANE.

DistilBERT is more than a head change. #208 added RoBERTa by reusing the shared BERT encoder graph and the _BERT_KEYS layer names, differing only host-side (head keys + position ids). DistilBERT's encoder names differ from BERT's, so it needs its own encoder key map too:

- Encoder layers are transformer.layer.{i}, not encoder.layer.{i}.
- Attention projections are q_lin / k_lin / v_lin / out_lin (not attention.self.query etc.); the FFN is ffn.lin1 / ffn.lin2; the norms are sa_layer_norm and output_layer_norm.
- No pooler and no token_type_embeddings; 6 layers.
- Head: pre_classifier -> ReLU -> classifier (note ReLU, not the two-layer tanh head BERT/RoBERTa use), on token 0.

Task:
- Add a DistilBERT encoder key map alongside _BERT_KEYS and route the layer-name prefix (transformer.layer vs encoder.layer) by family.
- Add the DistilBERT head to _seqcls_head (pre_classifier -> ReLU -> classifier, no pooler). The missing token_type table already falls back to zeros (#208).
- Detect the family by model_type == "distilbert" (or by head/key shape, matching how #208 detects RoBERTa vs BERT).

Acceptance:
- A DistilBERT AutoModelForSequenceClassification reranker runs on the ANE; scores match HF (small relerr) with ranking preserved.
- Off-device coverage for the head selection and the encoder key-map selection, in the style of tests/test_cross_encoder.py.

Pointers: aneforge/models.py (_BERT_KEYS, Encoder._build, CrossEncoder, _seqcls_head, _position_ids), #208 (the RoBERTa addition to model the structure on).

JulianZJN:
Closes #219

## Summary

- Add the DistilBERT encoder key map and transformer.layer.{i} prefix.
- Keep the zero token-type fallback and BERT-style position indexing for DistilBERT.
- Support the pre_classifier -> ReLU -> classifier head while preserving BERT/RoBERTa tanh behavior.
- Add off-device coverage for key selection and activation routing.

## Validation

- ruff, pylint (10.00/10), pyright, and compileall pass.
- pytest -m not requires_ane -q — 353 passed, 1 skipped, 697 deselected.
- PYTHONPATH=. uv run python tests/run_corpus.py — 90/90 variants green.
- On-device comparison with ml6team/cross-encoder-mmarco-german-distilbert-base on Apple M4 / macOS 26.5.1: max relative logit error 1.61%, ranking order matched.
- Full tests/ run: 1049 passed, 1 skipped; one unrelated abs VJP boundary miss (0.968 vs 0.97) in that run, with the targeted test passing on immediate rerun and five fixed hash seeds.

### DIFF
--- a/aneforge/models.py
+++ b/aneforge/models.py
@@ -271,6 +271,14 @@ def _seqcls_activation(values: np.ndarray, model_type: str) -> np.ndarray:
   return np.maximum(values, 0) if model_type == "distilbert" else np.tanh(values)
 
 
+def _token_type_ids(enc: dict, length: int, has_token_type_embeddings: bool) -> np.ndarray:
+  """Return segment ids, ignoring tokenizer output when the model has no segment table."""
+  if not has_token_type_embeddings:
+    return np.zeros(length, dtype=np.int64)
+  values = enc.get("token_type_ids")
+  return np.asarray(values if values is not None else [0] * length, dtype=np.int64)
+
+
 class CrossEncoder:
   """Reranker: scores (query, passage) pairs with a BERT-, RoBERTa-, or DistilBERT-family sequence-classification
   model, the transformer running on the ANE. Mirrors `sentence_transformers.CrossEncoder`:
@@ -308,7 +316,8 @@ class CrossEncoder:
     self.word = g(f"{pref}.embeddings.word_embeddings.weight")
     self.pos = g(f"{pref}.embeddings.position_embeddings.weight")
     typ_key = f"{pref}.embeddings.token_type_embeddings.weight"
-    self.typ = g(typ_key) if typ_key in sd else np.zeros((1, self.D), np.float32)
+    self._has_typ = typ_key in sd
+    self.typ = g(typ_key) if self._has_typ else np.zeros((1, self.D), np.float32)
     self.eln_w, self.eln_b = g(f"{pref}.embeddings.LayerNorm.weight"), g(f"{pref}.embeddings.LayerNorm.bias")
     self.layers = [{k: g(f"{pref}.{layer_prefix.format(i=i)}" + v) for k, v in layer_keys.items()}
                    for i in range(self.L)]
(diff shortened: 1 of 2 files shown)


sbryngelson:
Thanks for this -- the encoder key map, the head detection, and the ReLU-vs-tanh routing all look right, and the off-device tests are good. Nice work for a first PR.

I tested it on-device (M5 Pro) with a common DistilBERT reranker and hit a crash, so this needs one fix before it can merge:

```python
from aneforge.sentence_transformers import CrossEncoder
CrossEncoder("distilbert-base-uncased-finetuned-sst-2-english").predict([("a great film", "loved it")])
# IndexError: index 1 is out of bounds for axis 0 with size 1
```

Root cause: DistilBERT has no `token_type_embeddings`, so `CrossEncoder.__init__` falls back to a single zero row (`self.typ = np.zeros((1, D))`, models.py:311). But that model's tokenizer emits `token_type_ids = [0,0,0,0,0,1,1,1]` for a (query, passage) pair (its `model_input_names` includes `token_type_ids`). So `_embed` does `self.typ[typ_ids]` -> `self.typ[1]` -> IndexError (models.py:330). Your on-device model (`ml6team/...distilbert...`) happens to have a tokenizer that doesn't emit `token_type_ids`, which is why it passed for you.

BERT doesn't hit this (real 2-row table); RoBERTa/XLM-R has a 1-row table but its tokenizer emits no `token_type_ids`. DistilBERT is the case where a 1-row table meets a tokenizer that does emit them.

Suggested fix: when there is no real token-type table, the token-type term must be zero regardless of what the tokenizer returns. E.g. record `self._has_typ = typ_key in sd` in `__init__`, then in `predict` (models.py:343) use `typ = np.asarray(enc.get("token_type_ids") if self._has_typ else None or [0] * len(ids), dtype=np.int64)` -- or simplest, force zeros when `len(self.typ) == 1`. A regression test with a tokenizer that emits `token_type_ids` for a DistilBERT config would lock it in.

Happy to re-test on-device once it's updated.

sbryngelson:
Re-tested on an M5 Pro: the model that crashed before (distilbert-base-uncased-finetuned-sst-2-english) now runs on the ANE at relerr 4e-4 vs HF with ranking preserved, and your two new `_token_type_ids` tests lock in the fix. `_token_type_ids` forcing zeros when there's no embedding table is exactly right. Nice work turning this around for a first PR. Merging.

PR #224, opened 2026-08-15
GPT-2 decode: resident KV cache + long context (fold onto the LLM runner) (#215)

### LINKED ISSUE #215: GPT-2 decode: resident KV cache + long context (fold onto the LLM runner)
## Gap

#210 brought up GPT-2 as a **correctness demo**, not a usable generation path. The `GPT2` class in `aneforge/models.py` is separate from the mature `aneforge/llm.py` runner and shares none of its decode machinery:

- **No KV cache.** It recompiles the forward on the growing sequence (cached per length). Measured **0.35 tok/s** on an M5 Pro (the Qwen runner does ~75 tok/s for a similar size). O(S^2) recompute.
- **Native SDPA caps it at S < 512**, so long-context GPT-2 is unsupported. The runner already has a >512 attention path (`tests/test_llm.py::test_attention_context_above_512`).
- No segmentation past the ~2 GB program ceiling, no int8-decode, no speculative decode either.

The forward and greedy decode **are** validated (cosine 1.0000, 16/16 token match vs HF), so this is about making generation *fast and unbounded*, not correct.

## The underlying capability

The `llm.py` runner (resident-KV decode via `share_buffer`, >512 attention, segmentation, int8, spec-decode) is RMSNorm + RoPE only. GPT-2 needs three things it doesn't have: **LayerNorm** (affine, not RMSNorm), **learned positional embeddings** (a `wpe` table indexed by absolute position, vs RoPE), and a **tied lm_head**. So the real gap is "the runner can't do LayerNorm / learned-positional decoders" -- GPT-2 is just the first consumer.

## Two approaches (pick in the PR)

- **(A) Give the `GPT2` class its own resident-KV decode.** Contained, GPT-2-only: a decode graph that takes one token + the resident K/V, adds the positional embedding for the current length, and reads the resident cache under causal attention. Doesn't touch the runner. Less reuse.
- **(B) Generalize the runner** to support LayerNorm + learned-positional + tied-head via `LayerSpec`/an adapter, and route GPT-2 through `from_pretrained`. Unifying -- GPT-2 then inherits segmentation, int8, and spec-decode for free -- but touches the mature runner, so higher risk.

Recommendation: (B) if the runner's `LayerSpec` abstraction extends cleanly; (A) as a fallback if the norm/positional differences prove too invasive.

## Acceptance

- GPT-2 decode with a resident KV cache; tok/s within ~2x of a same-parameter Qwen on the same machine.
- Context beyond 512 tokens works (no S < 512 cap).
- Greedy decode still token-identical to Hugging Face fp32 (the #210 check must keep passing).

## Pointers

- `aneforge/models.py`: `class GPT2`, `load_gpt2`, `_gpt2_layers`
- `aneforge/llm.py`: the resident-KV decode path, `LayerSpec`, `_attn_decode`, `share_buffer` usage
- #210 (bring-up), #203 (the capability), `examples/gpt2.py`, `docs/developer/models.md` (load_gpt2 section)

Priority is moderate -- GPT-2 is semi-legacy, but the LayerNorm-decoder capability generalizes (encoder-decoder heads, other LayerNorm LMs).

axiom-of-choice:
Resolves #215

This PR implements the **Approach (B)** recommended in the issue: generalizing the mature LLM runner (`aneforge/llm.py`) and adapting the GPT-2 loader to use it.

By leveraging `LlamaPrefill`, GPT-2 gains access to:
- A resident KV cache (avoiding the prior O(S^2) recompute-per-step penalty)
- Memory segmentation (chunking layers automatically to fit under the 2 GB program cap)
- Long context windows (breaking the native SDPA 512-token cap limit)

### Changes
1. **Generalization (`llm.py`)**: 
   - Extended `LlamaConfig` with `norm_type` (supports `"rms"` or `"layer"`) and `rope` (boolean).
   - Modified `_attn_prefill`, `_attn_decode`, and `_attn_prefill_kv` to conditionally apply LayerNorm vs RMSNorm and bypass RoPE.
   - Added `_gelu_mlp` mapping using the existing `gelu_new` approximation.
   - Added `_gpt2_adapter` to parse the HF config/state_dict and map Conv1D weights (transposed for linear layers) and biases to the `llm.py` naming conventions. Registered it in `ADAPTERS`.
   - Included host-side sum of the absolute positional embedding (`wpe`) during prefill and per-step decode.
2. **Lowering (`_compile.py`)**:
   - Refactored the MIL lowering for `layer_norm` to operate natively on 2D arrays (`[M, D]`), avoiding the previous `reshape` to `[M, D, 1, 1]` that constrained sequence lengths to  \le 512$ due to Apple Neural Engine tensor limits.
3. **Refactor (`models.py`)**:
   - `GPT2` (and `load_gpt2`) now instantiates `LlamaPrefill` underneath and delegates the `__call__` and `generate` methods.
4. **Validation (`examples/gpt2.py`)**:
   - Updated the example script to test prefill cosine, greedy decode with the resident KV cache, and a batched >512-token prompt test.

### On-Device Verification (M2 Pro, A15 ANE)
- `examples/gpt2.py` passes perfectly:
  - **Prefill cosine vs HF fp32**: `1.00000` (exact top-1 match).
  - **KV Cache Decode**: 16/16 token-identical match with HF. Decode speed ~1.47 tok/s.
  - **Long Context**: Generated tokens at context >512 tokens flawlessly with batched prefill over a 520-token prompt.
- **Corpus**: `tests/run_corpus.py` (90/90 passes on device, `GATE: GREEN`). 

Code follows house style (2-space, ASCII, clean pyright/ruff).


aneforge/llm.py
```diff
@@ -533,15 +595,42 @@ def _gpt2_adapter(c, sd) -> tuple[LlamaConfig, dict]:
 # wins, dense Llama/Qwen is the fallback. New archs append here (e.g. aneforge.moe) so the loader stays generic.
 ADAPTERS: list = [
   (lambda c: getattr(c, "model_type", None) == "gpt2", _gpt2_adapter),
```
axiom-of-choice: Maybe decouple "Gemma" "mistral" etc into an enum

aneforge/llm.py
```diff
@@ -186,24 +200,29 @@ def _attn_decode(x: Tensor, w: dict, cfg: LlamaConfig, ls: LayerSpec, ctx: dict,
...
   Kout = Kin * inv + k * oh; Vout = Vin * inv + v * oh        # masked positional write into the cache
   Kr, Vr = _repeat_kv3(Kout, H // KV), _repeat_kv3(Vout, H // KV)
   sc = ((q @ Kr.transpose([0, 2, 1])) * (1.0 / dh ** 0.5) + mask).softmax(-1)   # [H, 1, M]
   a = (sc @ Vr).reshape(H, 1, dh).transpose([1, 0, 2]).reshape(1, H * dh)
-  return x + a.linear(w["wo"]), [(Kout, Kin), (Vout, Vin)]
+  return x + a.linear(w["wo"], w.get("o_bias")), [(Kout, Kin), (Vout, Vin)]
 
 
 PREFILL_MIXERS = {"attention": _attn_prefill}
-PREFILL_MLPS = {"swiglu": _swiglu, "geglu": _geglu}
```
axiom-of-choice, line 204: I don't like this constants buryd in the code but I think we cannot move it to the top as they reference functions defined afterwards. Double check.

aneforge/models.py
```diff
@@ -454,6 +454,8 @@ def _gelu_new(x: Tensor) -> Tensor:
   return (x * 0.5) * inner.tanh().adds(1.0)
 
 
+# smallest supported matmul output dim on family 3 (A13-A15, M1/M2); see bench/decode_measurement.py:130-135
+_LMHEAD_TILE = 16384
```
axiom-of-choice: Confirm this, and if we need to add more constants per family.

aneforge/models.py
```diff
@@ -474,89 +476,60 @@ def _logits_from(net: MultiModel | Model, out) -> np.ndarray:
...
+  `.generate(prompt, K)` autoregressively generates K tokens using the resident KV cache."""
 
   def __init__(self, name: str, int8: bool = False, max_layers: int | None = None) -> None:
     from transformers import AutoConfig, AutoModelForCausalLM, AutoTokenizer  # lazy
-    cfg = AutoConfig.from_pretrained(name)
-    if cfg.model_type != "gpt2":
-      raise ValueError(f"load_gpt2: {name!r} is not a GPT-2 model (model_type={cfg.model_type!r})")
+    from .llm import _gpt2_adapter, LlamaPrefill
+    cfg_hf = AutoConfig.from_pretrained(name)
+    if cfg_hf.model_type != "gpt2":
```
axiom-of-choice: Again enum could be useful

axiom-of-choice:
Replaced manual strings with `ModelType` enums across the tests in `test_dense_loader_guard.py` and the mappings in `llm.py` as discovered
I have rebased and all tests including `tests/test_layer_norm_affine_form.py` are now perfectly aligned with the new 2D implementation.

aneforge/llm.py
```diff
@@ -186,24 +200,29 @@ def _attn_decode(x: Tensor, w: dict, cfg: LlamaConfig, ls: LayerSpec, ctx: dict,
...
+PREFILL_MLPS = {"swiglu": _swiglu, "geglu": _geglu, "gelu_new": _gelu_mlp}
```
axiom-of-choice, line 241: I don't like this constants buryd in the code but I think we cannot move it to the top as they reference functions defined afterwards. Double check.

aneforge/llm.py
axiom-of-choice: Checked: in Python, functions like `_attn_prefill`, `_swiglu`, `_geglu`, `_gelu_mlp`, and `_attn_decode` must be defined before populating `PREFILL_MIXERS` / `PREFILL_MLPS` with direct function references. 

Placing the registry dictionaries right after their builder functions and immediately preceding `LlamaPrefill` (which consumes them) is standard and idiomatic in the codebase (modules like `moe.py` also import and extend them). Moving them to the top would require forward references or lambda wrappers which would add runtime overhead and break static typing.

aneforge/llm.py
```diff
@@ -519,16 +558,62 @@ def _cfg_from_hf(c) -> LlamaConfig:
...
+
+class ModelType(str, Enum):
+  GPT2 = "gpt2"
+  GEMMA = "gemma"
+  GEMMA2 = "gemma2"
+  GEMMA3 = "gemma3"
+  GEMMA3_TEXT = "gemma3_text"
+  MISTRAL = "mistral"
+  LLAMA = "llama"
+  QWEN = "qwen"
```
axiom-of-choice: This is buried, let's move it to the top or other compatible file.

aneforge/llm.py
```diff
@@ -519,16 +558,62 @@ def _cfg_from_hf(c) -> LlamaConfig:
...
-  (lambda c: c.model_type == "mistral", _mistral_adapter),
+  (lambda c: getattr(c, "model_type", None) == ModelType.GPT2, _gpt2_adapter),
+  (lambda c: getattr(c, "model_type", None) == ModelType.GEMMA, _gemma_adapter),
+  (lambda c: getattr(c, "model_type", None) == ModelType.MISTRAL, _mistral_adapter),
 ]
 
 # Gemma-2/3 reach the dense fallback (Llama-shaped) but need softcapping / GeGLU / (1+w)-norm it lacks;
 # a name backstop for the softcapping feature check in _assert_dense_compatible.
-_DENSE_INCOMPATIBLE = {"gemma2", "gemma3", "gemma3_text"}
+_DENSE_INCOMPATIBLE = {ModelType.GEMMA2, ModelType.GEMMA3, ModelType.GEMMA3_TEXT}
```
axiom-of-choice: This is also kinda buried, we should move it to the top...

axiom-of-choice (commented):
Replaced manual strings with `ModelType` enums across the tests in `test_dense_loader_guard.py` and the mappings in `llm.py` as discovered
I have rebased and all tests including `tests/test_layer_norm_affine_form.py` are now perfectly aligned with the new 2D implementation.

tests/test_gpt2.py
```diff
@@ -84,6 +85,60 @@ def test_gpt2_tiled_head_lowers():
...
+    model._hidden(np.arange(11, dtype=np.int64))
+  with pytest.raises(ValueError, match="exceeds this model's 10 max positions"):
+    model.generate([1, 2, 3], max_new_tokens=4, max_len=20)
+
+
+def test_llama_prefill_release_clears_state():
+  """`release()` nulls `_net`/`_dec`/`_pre` (not just releasing the underlying program), so a
+  subsequent call recompiles instead of replaying a freed program."""
+  class FakeConfig:
+    model_type = "gpt2"
```
axiom-of-choice: Why not import the enum and use it?

sbryngelson:
Tested end-to-end on an M5 Pro. examples/gpt2.py: prefill cosine 1.00000 vs HF, resident KV-cache decode greedy 16/16 match, and the >512-token path works (520-token prompt prefilled + decoded). Decode is 508 ms/token (1.97 tok/s) vs main's 2839 ms/token (0.35) -- the KV cache is a real ~5.6x. Regressions all green: test_llm.py 13/13 on the ANE (existing Llama/Qwen/Gemma/Mistral unaffected by the norm_type/rope generalization), the layer_norm affine guards 17/17 (the 2D layer_norm rewrite keeps the #201/#204 invariant), off-device suite 349 passed.

This delivers #215's core (resident KV cache, long context, correct output), so merging.

One follow-up worth tracking: 1.97 tok/s is still well below a same-size Qwen (~75 tok/s). Likely the per-step tiled lm_head (4 dispatches) and the host-side wpe add each step rather than the layer compute. Not a blocker (correct, and much faster than before), but if you agree I'll file a throughput follow-up. Nice work landing Approach B.

axiom-of-choice:
>  lm_head (4 dispatches) and the host-side wpe add each step rather than the layer compute. Not a blocker (correct, and much faster than before), but if you agree I'll file a throughput follow-up. Nice work landing Approach B.

Yes, let's file a ticket and follow up work on optimizing throughput

sbryngelson:
Filed as #226. Two concrete levers there: family-aware lm_head tiling (the head tiles at 16384 unconditionally, but 50257 fits one tile on A16+, so 4 dispatches -> 1 on M5/M4) and folding the wpe add into the resident decode graph to drop the per-step host round-trip -- with a profile-first step to confirm those are the cost.

Issue #227, opened 2026-08-16
Decode throughput: move lm_head to ANE for all LlamaPrefill models


axiom-of-choice:
#215 / #224 folded GPT-2 into LlamaPrefill with host-side lm_head and wpe. speculative.py already has _build_lm_head that runs the head on the ANE as a tiled compile_multi, amortising cost over Kq tokens. That pattern should be generalised to LlamaPrefill.generate so every model (Llama/Qwen/GPT-2/Mistral) benefits, not just the speculative verifier.
Scope:
- Add _build_decode_lm_head to LlamaPrefill, using limit("max_tensor_dim", detect_family()) so A16+ models with vocab < 65536 get a single dispatch.
- Replace _logits host-matmul in generate with ANE lm_head when the model opts in (or by default if numerics stay exact).
- Ensure release() frees the lm_head program.
Blocked on #226 completing first, so we have the family-aware tile size and a measured GPT-2 baseline to compare against.

axiom-of-choice:
Will take this after i finish with #226

PR #229, opened 2026-08-16
perf: GPT-2 decode throughput (profile + wpe fold + host lm_head fp32 cache)

### LINKED ISSUE #226: GPT-2 decode throughput: family-aware lm_head tiling + fold wpe into the graph
#224 gave GPT-2 a resident KV cache, so decode went from 0.35 tok/s (recompute-per-length) to 1.97 tok/s on an M5 Pro (508 ms/token, gpt2-medium). Correct (greedy 16/16 vs HF), but still well below a same-size Qwen (~75 tok/s), so #215's "within ~2x of a same-size Qwen" bar is not met yet. This tracks the throughput work.

Two concrete levers, plus a measurement step:

1. Family-aware lm_head tile size. `_lm_head_tiles` (aneforge/models.py:457) tiles the tied head at a hardcoded `_LMHEAD_TILE = 16384`, which is the A13-A15 cap. On A16+ (M4/M5) the per-op cap is 65536, and GPT-2's 50257 vocab fits a single tile there -- so the head runs as 4 separate ANE dispatches per token when it could be 1. Make the tile size family-aware (65536 on A16+) so the head is one dispatch on current silicon.

2. Fold the wpe positional add into the resident decode graph. The learned positional embedding (`self.wpe`, models.py:506) is added host-side each decode step, which is a host round-trip per token. The token embedding already uses a position one-hot into the graph; do the same for wpe so the per-step add stays on the engine.

3. Profile before optimizing. Measure where the 508 ms/token actually goes -- the layer program vs the head vs the host-side embedding add -- so the two changes above are confirmed to be the cost, not assumed. A per-stage timing print in examples/gpt2.py's decode loop is enough.

Acceptance:
- GPT-2 decode throughput within ~2x of a same-parameter Qwen on the same machine.
- Greedy decode still token-identical to HF (the #224 check must keep passing).

Pointers: aneforge/models.py (`_lm_head_tiles`, `_LMHEAD_TILE`, `GPT2.generate`, `self.wpe`), the resident-KV decode path in aneforge/llm.py, #224 (delivered the KV cache), #215 (parent).

axiom-of-choice:
READY:
Resolves #226 (throughput levers), plus a host-side `lm_head` fix surfaced by the profiling step the issue asked for.

## What's in this PR

1. **Per-stage decode profiling** (`examples/gpt2.py` + `on_stage` in `LlamaPrefill.generate`): splits per-token time into embedding / layers (ANE) / lm_head (host) / sample, so the cost is measured, not assumed (issue step 3).
2. **Family-aware `_lm_head_tiles`**: tile size comes from `_targets.limit("max_tensor_dim", detect_family())` instead of a hardcoded 16384. A16+ gets 65536 (GPT-2's 50257 vocab fits one tile); A13-A15 keep 4 tiles.
3. **Fold `wpe` into the resident decode graph**: the first decode chunk adds the learned positional embedding in-graph via a position one-hot + a baked `wpe` constant, removing the per-token host-side add.
4. **Host `lm_head` fp32-contiguous cache** (`LlamaPrefill._logits`): the weight is fp16 and NumPy has no fp16 BLAS, so every token re-converted a strided transpose (~124 ms). A contiguous fp32 transpose is now built once and reused (~10 ms), bit-identical to the previous per-call matmul.

## Measurements (M2 Pro, macOS 26.5.2, fp16)

GPT-2 medium, resident KV-cache decode (compile excluded):

| Stage | Before | After |
|---|---|---|
| host lm_head | 124.4 ms/token | 10.0 ms/token |
| ANE layers | 13.4 ms/token | 13.6 ms/token |
| decode (from stage timing) | ~136 ms/token (7.3 tok/s) | 23.8 ms/token (42 tok/s) |

Qwen3-0.6B, same fix, same machine: 2.1 → 16.6 tok/s (TPOT 480 → 62 ms). The host lm_head conversion dominated both, not the tiling.

## Honest accounting

- **Only lever 4 moved the needle.** The wpe fold is neutral for throughput (the host add it removed was microseconds) and cost ~0.2 ms/token in the ANE layers (13.4 → 13.6 ms; within run-to-run noise). The family-aware tile is correct but latent: `_lm_head_tiles` has no production call site today — its only caller (`GPT2.__call__`) was removed by #224 when GPT-2 folded onto `LlamaPrefill`, and `models.py` reaches it from tests only. It becomes live when #227 wires an ANE head back into decode.
- **The wpe fold is a numerics change, not just a move.** Decode now adds `wpe` in fp16 inside the graph, while prefill (`_hidden`/`_prefill_seed`) still adds it in fp32 on the host, so the same position is computed slightly differently per path. Greedy stays 16/16 identical to HF, but this is not "numerically identical" the way the lm_head cache is (that one is bit-exact).
- **`detect_family()` reads the host chip**, so a cross-compile for another family would tile for the host unless `ANEFORGE_TARGET` is set. Kept as the issue specified.

## Validation

- GPT-2: prefill cosine 1.00000, greedy 16/16 vs HF, long context >512 PASS.
- On-device `tests/test_gpt2.py` + `tests/test_llm.py`: 26 passed.
- Off-device suite: 330 passed; new `test_logits_caches_contiguous_fp32` (bit-identity + cache reuse), `test_gpt2_lm_head_tiles_follow_target_family`, and `warmup` position-guard coverage.
- Corpus gate: 90/90 GREEN.
- ruff, pylint 2-space gate, compileall clean.

## Review feedback applied

- `release()` now drops `_lmT` (the runner's largest host allocation, ~206 MB on gpt2-medium, ~2.5 GB on a 151936×4096 head), and the class docstring documents it.
- `warmup()` now hits the `_check_positions` guard via `_decoder` (previously a `warmup(2048)` on GPT-2 would silently truncate `wpe` and fail with a graph shape error instead of the clear ValueError).
- `on_stage` documented in `generate`'s docstring, including that `step` nests `embedding` + `layers` (summing all stages double-counts).

## Follow-ups

- #227: lm_head on the ANE for all LlamaPrefill models (tiled, speculative-style). This host fix is cheap and generic; #227 is the deeper optimization and should build on the `_lmT` cache.
- Want a stable repeated run and an A16 cross-check before leaving draft.

### DIFF
--- a/aneforge/llm.py
+++ b/aneforge/llm.py
@@ -8,6 +8,7 @@ the plan + canonical weights - not new branches in the hot path."""
 from __future__ import annotations
 from dataclasses import dataclass, field
 from enum import Enum
+import time
 import numpy as np
 
 from .graph import Tensor, input as _input, concat as _concat, _const
@@ -352,6 +353,10 @@ class LlamaPrefill:
       ctx = {"oh": _input((1, M, 1)), "inv": _input((1, M, 1)), "mask": _input((1, 1, M)),
              "cosp": _input((1, dh)), "sinp": _input((1, dh))}
       h = x; pairs = []
+      wpe_pos = None
+      if gi == 0 and "wpe" in self.w:
+        wpe_pos = _input((1, M))
+        h = h + wpe_pos @ _const(self.w["wpe"][:M].astype(np.float16))
       for li in grp:
         ls = self._spec(li); lw = self.w["layers"][li]
         h, sp = DECODE_MIXERS[ls.mixer](h, lw, cfg, ls, ctx, M)
@@ -365,6 +370,7 @@ class LlamaPrefill:
       for o, i in pairs:                                        # alias each state output -> its input (resident)
         net.prog.share_buffer(0, om[o], 0, inm[id(i)])
       p = {"x": inm[id(x)], "h": om[h], "states": {inm[id(i)]: i.shape for _, i in pairs}}
+      if wpe_pos is not None: p["wpe_pos"] = inm[id(wpe_pos)]
       for k, t in ctx.items():
         if id(t) in inm: p[k] = inm[id(t)]                      # only ports the chunk's mixers actually use
       chunks.append({"net": net, "p": p})
@@ -452,12 +458,15 @@ class LlamaPrefill:
         c["net"].prog.set_input(port, buf)
 
   def generate(self, token_ids, max_new_tokens=40, eos_id=None, max_len=None, on_token=None,
-               temperature=0.0, top_p=1.0, top_k=0, batched_prefill=True, prefill_pad=None):
+               temperature=0.0, top_p=1.0, top_k=0, batched_prefill=True, prefill_pad=None, on_stage=None):
     """Autoregressive generation with a resident KV-cache. The decode program (compiled once and cached) runs
     all layers for one token attending to caches that stay on the ANE across steps, so each step feeds only the
     token embedding + a position one-hot. `on_token(id)` is called per token (streaming). Greedy by default
     (`temperature=0`); pass temperature/top_p/top_k to sample. Returns the generated token ids."""
     cfg = self.cfg; f16 = np.float16
+    profiling = on_stage is not None
+    def stage(name, elapsed):
+      if on_stage is not None: on_stage(name, elapsed)
     prompt = [int(t) for t in token_ids]; M = int(max_len or len(prompt) + max_new_tokens)
     self._check_positions(M)
     eos = {int(eos_id)} if isinstance(eos_id, int) else {int(e) for e in (eos_id or ())}   # any-of stop set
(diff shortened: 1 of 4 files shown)


sbryngelson:
Tested on an M5 Pro. The core fix is a big, correct win: examples/gpt2.py decode is 7.1 ms/token (~140 tok/s, up from 1.97 with #224), lm_head down to 1.4 ms/token, cosine 1.00000, greedy 16/16. No LLM regression -- test_llm.py 13/13 on the ANE (the _logits fp32 cache and the check-positions change don't touch existing Llama/Qwen output). Good profiling call: the host fp16->fp32 lm_head reconversion was the real cost (and it was likely my #207 fp16 state-dict change that introduced it), so this speeds up all decoders, not just GPT-2.

One thing I had to fix to merge: test_gpt2_lm_head_tiles_shapes still had the hardcoded 4-tile/16384 assertion, so it FAILS on A16+ (this M5), where the family-aware _lm_head_tiles correctly returns one 50257 tile. Off-device CI passed it, so it would have merged green and broken every A16+ test run. Your new test_gpt2_lm_head_tiles_follow_target_family covers both families, but the old one wasn't updated -- I pushed a commit making it mock the cap so it covers both regimes too. Merging.

PR #231, opened 2026-08-18
examples: chat with your docs, entirely on the Neural Engine

sbryngelson:
An interactive "chat with your docs" demo where every model stage runs on the ANE: embeddings (all-MiniLM-L6-v2), reranking (ms-marco-MiniLM-L-6-v2), and generation (Qwen3-0.6B). Point it at a folder of .md/.txt and ask questions in a loop; it retrieves, reranks, and streams an answer, all on the engine. With no path it indexes this repo's own docs, so `python3 examples/rag_chat.py` runs immediately.

## What's here

- `examples/rag_chat.py` -- the CLI/REPL and the `Pipeline` (index -> retrieve -> rerank -> generate), per-stage latency, and an optional `--energy` flag that reports per-query joules via powermetrics (default runs are latency-only and need no sudo).
- `examples/_rag.py` -- the pure orchestration (token-window chunking, cosine top-k, budget-aware prompt packing), unit-tested off-device.
- `tests/test_rag.py` -- off-device unit tests plus one `requires_ane` end-to-end smoke test.
- Docs: a section in `docs/developer/models.md` and a README line.

## Two aneforge bugs this surfaced

Building over a real corpus exposed a systemic issue, so the demo works around it and I've filed a follow-up for the proper fix:

- The `Encoder` and `CrossEncoder` compile one ANE program per input sequence length and cannot pad (no attention mask). A 696-chunk corpus spans hundreds of distinct lengths and exhausts the engine (`op_create err=13`); padding to a fixed length corrupts embeddings (real tokens attend to `[PAD]`). The demo avoids it with uniform token-window chunks embedded via the public `st.encode` (verified cos 0.9999999 vs the reference), and by clearing the reranker's program cache per query. This affects any corpus embedding through the sentence-transformers path, not just this demo -- tracked separately for a core fix.

## Testing (M5 Pro)

- Full repo-docs index (786 chunks) builds with no crash; a multi-question session runs without exhausting the engine.
- Embeddings match the reference at cos 0.9999999; answers are clean and stop at EOS (the demo uses the model's chat template).
- `tests/test_rag.py`: 6 passed (5 off-device + the on-ANE smoke). ruff clean; mkdocs builds.

Decode is small-model quality (Qwen3-0.6B), and the demo works best on prose docs -- a folder of very many tiny files can still hit the per-length embedder limit (noted in the docs).

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,8 @@ model via `af.load_llm` and benchmarks prefill/decode (matching Hugging Face
 logits). For retrieval,
 [`examples/rag_embeddings.py`](examples/rag_embeddings.py) is a LangChain
 `Embeddings` drop-in backed by the on-ANE encoder (4-5x faster than the GPU,
-cosine 1.0000).
+cosine 1.0000). Chat with your docs entirely on the Neural Engine -- embeddings,
+reranker, and LLM on one engine: `python3 examples/rag_chat.py`.
 
 ## How it compares
 
--- a/docs/developer/models.md
+++ b/docs/developer/models.md
@@ -204,3 +204,9 @@ call and reports the query's energy as `~NNN mJ this query, 0 GPU` -- 0 GPU beca
 whole pipeline, including generation, never leaves the Neural Engine. Without `sudo` (or
 without `powermetrics` on the machine), the flag falls back to running the query with no
 energy figure rather than failing.
+
+The demo works best on prose documents. Chunking windows to a fixed token count, but a
+file shorter than that window still embeds as a single chunk sized to its own length, so
+a folder of very many tiny files (each a different length) can still push the ANE
+embedder toward compiling one program per distinct sequence length -- the same per-length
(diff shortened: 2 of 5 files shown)


sbryngelson:
The core aneforge bug this surfaced (Encoder/CrossEncoder compile one program per input length, no attention-masked padding) is filed as #232 for a proper fix. This PR's uniform-chunk + cache-clear approach is the demo-side workaround.

sbryngelson:
Updated on top of the merged #233 (Encoder/CrossEncoder padding fix). The two workarounds this PR needed are now gone: plain character chunking (was token-window), no per-query reranker cache-clearing, TOP_K back to 20. On an M5 Pro the fix cuts index build 79s -> 21s and rerank ~6.1s -> ~0.78s/query, same correct source-grounded answers, no crash. The demo now demonstrates the clean path rather than papering over the bug.

PR #240, opened 2026-08-18
perf: tiled ANE lm_head for LlamaPrefill decode (#227)

axiom-of-choice:
## What's in this PR

Opt-in tiled lm_head on the ANE for `LlamaPrefill`, replacing the host fp32 matmul. `compile_multi`
tiles the vocab axis to the target family's `max_tensor_dim` (16384 on A13-A15, 65536 on A16+), so
every tile ships in one program and costs one dispatch per token.

- `_lm_head_tile_size()`, `_lm_head_program()` (built once, cached, independent of `max_len`), `_ane_logits()`
- `_logits(hidden_row, ane=None)` routes to the ANE head; `ane` is the per-call override, threaded down from `generate`
- `release()` frees the program, `warmup()` compiles it, `af.load_llm(..., ane_lm_head=True)` enables it
- Declines to the host path on opt-out, a head over `_lmhead_bytes` (1.2 GB), or a failed compile: warn once, never break `generate`

Default is `False`.

## Measurements

M2 Pro, macOS 26.5.2, fp16, family 3 (cap 16384). Interleaved host/ANE rounds in one process, both
paths pre-warmed, median of 3 runs, first token dropped.

| model | tiles | lm_head host | lm_head ANE | decode host | decode ANE |
|---|---|---|---|---|---|
| gpt2-medium (vocab 50257) | 4 | 10.8 ms/tok | 3.4 ms/tok | 39.6 tok/s | 59.4 tok/s |
| Qwen3-0.6B (vocab 151936) | 10 | 30.1 ms/tok | 8.2 ms/tok | 17.8 tok/s | 29.3 tok/s |

`layers` is unchanged (13.9 vs 13.5, and 25.9 vs 25.9 ms/tok), so the whole delta is the head. Ten
output ports do not eat the gain: one program, one dispatch, so the cost tracks bytes and not tiles.
Greedy output identical host vs ANE, 64/64 tokens, both models. `_lmT` is never allocated on the ANE
path: 206 MB of host memory on GPT-2, 622 MB on Qwen3-0.6B. Head compile is 0.4 s and 1.8 s, paid by
`warmup`.

GPT-2 medium vs Hugging Face with the ANE head serving prefill, seed and decode: prefill cosine
1.00000, argmax match, 16/16 greedy tokens.

## Honest accounting

- Not bit-exact. The hidden row casts to fp16 and the matmul runs in fp16. Greedy matched on both
  models measured, but argmax can flip on a close tie, so the default stays `False`.
- The head compiles fp16 even when `compress="int8"`: quantizing the head has no residual to absorb
  the error. Conservative choice, not measured.
- Tied heads pay twice. In GPT-2 `lm_head is embed`, so the ANE program bakes 103 MB while `embed`
  stays on the host for the token gather. The `_lmT` saving is real, the device cost is not free.
- The A16 single-tile branch is covered off-device only (monkeypatched family). Not run on hardware.
- `compile_multi` never calls `_retarget_for`, so the multi-output path has no oversize gate. The
  tile size is the only thing keeping tiles legal, and pointing `ANEFORGE_TARGET` above the host
  family will tile for a cap the host cannot run.
- `mkdocs build --strict` not run: mkdocs is not installed locally. `docs/llm.md` only adds prose and
  a table, and `tables` is already enabled in `mkdocs.yml`.

## Validation

Off-device: `ruff check` clean, pylint (W0311/C0303) 10.00, `pyright aneforge` no new errors vs
`main` (the 22 it reports are missing optional imports, identical on both), `compileall`,
`pytest -m "not requires_ane"` 356 passed / 6 skipped.

On-device (M2 Pro): corpus `GATE: GREEN` 90/90. `pytest tests/` 1045 passed, 13 skipped, 4 failed;
the 4 are missing optional deps in my environment (`torchvision` for the ONNX ResNet-18 and CIFAR
tests, `scipy` for the two `erf` tests), unrelated to this change.

Ten off-device tests: per-family tile widths (not just the count), oversize decline,
compile-failure fallback, dim-mismatch fallback, `release()`, decoder rebuild not invalidating the
program, program built once, host path bit-identical when off, per-call override not touching
instance state, and the override not leaking out of `generate`.

Two on-device tests: ANE vs host logits over a 3-tile head (argmax plus fp16 tolerance), and greedy
`generate` parity host vs ANE with `_lmT` never built.

## Follow-ups

Found while working on this, not touched here:

1. `speculative._build_lm_head` hardcodes 65536 (`speculative.py:72`). The real cap is 16384 on
   A13-A15 and `compile_multi` has no gate, so `spec_generate` with a large vocab on an M1/M2/M3
   fails without a clear error. Current tests use vocab 48 and miss it.
2. `LlamaPrefill.release()` does not free `_verifier`, its chunks, or its lm_head. Preexisting leak
   from speculative decoding.
3. Four implementations of a tiled lm_head now: `models._lm_head_tiles` (no call site),
   `speculative._build_lm_head`, `bench/decode_measurement._lm_head` (16384 fixed), and this one.
4. `moe.load_gguf` keeps `lm_head` fp32 on the host for exactly the reason this PR addresses, so it
   could drop that allocation too.

### DIFF
--- a/aneforge/llm.py
+++ b/aneforge/llm.py
@@ -6,6 +6,7 @@ The runner is model-agnostic: a `LlamaConfig` carries a per-layer plan (`LayerSp
 and an MLP from the builder registries below, so supporting a new architecture is an *adapter* that emits
 the plan + canonical weights - not new branches in the hot path."""
 from __future__ import annotations
+import warnings
 from dataclasses import dataclass, field
 from enum import Enum
 import time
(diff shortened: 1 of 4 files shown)


sbryngelson (approved):
Audited on an M5 Pro (family 5, cap 65536), both tile regimes.

Off-device tests/test_lm_head.py pass 10/10 and the on-device tests pass 2/2. mkdocs build --strict is clean here, which resolves the caveat about not running it locally.

The single-tile path works on real silicon (the A16 branch you could only monkeypatch): gpt2 vocab 50257 is one tile, greedy is identical host vs ANE over 24 tokens, _lmT is never built on the ANE path, and logits are cosine 1.00000 vs HF with a matching argmax. The multi-tile path (Qwen3-0.6B, 3 tiles at cap 65536) gives logits cosine 1.00000 vs host with a matching first-token argmax, so the concatenation is correct. The greedy divergence I saw on Qwen over 20 tokens is a genuine near-tie flip (host top-2 gap 0.005, fp16 head error 0.005, cosine 1.0), not a concat bug, which is the fp16 behavior you document for defaulting to False.

The tiling, host-side concat, opt-out/oversize/compile-fail fallback, per-call override threading, release(), and built-once caching are all correct and well covered.

Three non-blocking notes:

1. Execute-time failures aren't caught. The fallback guards compile (_lm_head_program), but _ane_logits calls pr.execute() unguarded, so a head that compiles but fails at execute (the ANEFORGE_TARGET-above-host-family case you flagged, if compile doesn't reject the tiles) would crash generate instead of falling back. Wrapping the first execute in a try that declines to host would close it.

2. The on-device exact-match tests are seed-fragile. test_generate_ane_lm_head_matches_host and the argmax half of test_ane_lm_head_matches_host_logits assert exact fp16-vs-fp32 equality on random-weight models, and near-ties flip, as the Qwen run shows. seed=3 happens not to flip on the vocab-200 model, but a future RNG-stream change could. requires_ane only, so no CI impact. The rel < 5e-3 check is the robust one.

3. A dim mismatch is reported as "compile failed (...)" in the warning, though nothing was compiled.

Correct on real silicon in both regimes. The notes are minor.

PR #249, opened 2026-08-24
warn on compress= fallback to wider encoding

### LINKED ISSUE #222: compress='int4' falls back to int8 silently: no way to see the applied weight encoding
### What happens

`compress="int4"` is accuracy-gated per weight, and a rejected weight falls back to int8 (or fp16). The docstring says so. What is missing is any way to observe which encoding was actually applied: no warning, no return value, nothing on the returned `Model`. A caller who asks for int4 and gets int8 has no runtime signal at all.

### Reproducer

```python
import os, tempfile, numpy as np, aneforge as af
np.random.seed(0)
W = np.random.randn(512, 512).astype(np.float16) * 0.04

def blob_and_op(**kw):
    d = tempfile.mkdtemp()
    x = af.input((1, 512)); y = x.linear(W)
    af.compile(y, build_dir=d, **kw)
    b = sum(os.path.getsize(os.path.join(r, f))
            for r, _, fs in os.walk(d) for f in fs if f.endswith(".bin"))
    mil = [os.path.join(r, f) for r, _, fs in os.walk(d) for f in fs if f.endswith(".mil")]
    t = open(mil[0]).read()
    op = next((o for o in ("constexpr_lut_to_dense", "constexpr_affine_dequantize")
               if o in t), "const (fp16)")
    return b, op

for label, kw in [("compress='int4' (default atol=0.05)", dict(compress="int4")),
                  ("int8=True",                          dict(int8=True)),
                  ("compress='int4', compress_atol=0.15", dict(compress="int4", compress_atol=0.15))]:
    b, op = blob_and_op(**kw)
    print(f"{label:38s} -> weights {b:7d} B   {op}")
```

```
compress='int4' (default atol=0.05)    -> weights  263360 B   constexpr_affine_dequantize
int8=True                              -> weights  263360 B   constexpr_affine_dequantize
compress='int4', compress_atol=0.15    -> weights  131296 B   constexpr_lut_to_dense
```

The first two rows are byte-identical. `compress="int4"` produced an int8 program.

### Why it is worth a signal

On a ~600 M-parameter transformer decoder the two spellings differ by a lot more than a flag:

| requested | weight blob | emitted op | tok/s at B=1 |
|---|---|---|---|
| `compress="int4"` (default atol) | 575.0 MB | `affine_dequantize` | 123.2 |
| `compress="int4", compress_atol=0.15` | 287.3 MB | `lut_to_dense` | 203.1 |

A 1.65x throughput difference between what was asked for and what was applied, with nothing in the API to distinguish them. The only way I found to tell was passing `build_dir=`, stat-ing `weights.bin`, and grepping the emitted MIL for the constexpr op. Anyone benchmarking encodings, or reporting int4 numbers, can silently measure int8 instead.

`compress="auto"` makes it harder still, since the choice is per weight and the mix is invisible.

### Suggestion

Either would fix it, and they compose:

1. **Warn on fallback**, following the existing `DispatchFloorWarning` precedent, e.g. `aneforge.CompressionFallbackWarning: compress='int4' rejected for 47/48 weights (relative L2 > compress_atol=0.05); applied int8`. Cheap, and it lands where the mistake happens.
2. **Expose the applied encoding** on the returned `Model`, e.g. `model.weight_encodings -> {"int4-LUT": 1, "int8": 47}` plus the streamed byte total. Useful for `compress="auto"`, and it makes the encoding assertable in a test.

Happy to send a PR for (1) if you would take it.

axiom-of-choice:
Closes #222.

When `compress='int4'` (or `'sparse'`/`'blockwise'`) rejects a weight because its relative error exceeds `compress_atol`, the weight silently falls through to int8 or fp16. A caller who asked for int4 and got int8 has no runtime signal at all.

## What this does

Adds `CompressionFallbackWarning`, emitted after the emit loop when any requested-compression weight was rejected. The message reports:

- How many weights were rejected (e.g. `47/48`)
- Which encoding was requested and which was applied
- How to widen the tolerance (`compress_atol`)

Follows the existing `DispatchFloorWarning` precedent. Filterable with `warnings.filterwarnings`.

## Changes

- `_Emitter`: tracks `_n_weights` and `_n_rejected` dict; each fallthrough point (sparse, int4, blockwise) increments its counter
- `CompressionFallbackWarning` class + `_compression_fallback_signal()` called from `compile()`, `compile_multi()`, `_compile_segmented()`
- 5 tests: warn on reject, no warn on accept, no warn on `compress=None`, filterability, mixed accept/reject counts

## Checks

- `ruff check` clean
- `pyright aneforge` 0 errors
- `pylint` 10.00/10
- `pytest -m 'not requires_ane'` 383 passed, 2 skipped

### DIFF
--- a/aneforge/__init__.py
+++ b/aneforge/__init__.py
@@ -22,7 +22,7 @@ from .graph import (Tensor, affine, batch_norm, batch_to_space, channel_to_space
                     space_to_channel, space_to_depth, split, sort, stack, topk,
                     upsample_bilinear, where)
 from ._compile import (Model, SegmentedModel, compile, PrecisionWarning,
-                       CrossChipFP16Warning, DispatchFloorWarning)
+                       CrossChipFP16Warning, DispatchFloorWarning, CompressionFallbackWarning)
 from ._paired import Paired, paired
 from ._optimize import tune, tune_precision, tune_attention, attention_tiles
 from ._cost import estimate, estimate_provenance, precision_risk, project_peak
@@ -39,7 +39,7 @@ from . import moe as moe   # registers the "moe" MLP + Qwen3-MoE adapter into th
 
 __all__ = [
     "Tensor", "Model", "SegmentedModel", "PrecisionWarning", "CrossChipFP16Warning",
-    "DispatchFloorWarning",
+    "DispatchFloorWarning", "CompressionFallbackWarning",
     "input", "image_input", "conv", "conv_transpose", "dynamic_conv", "concat",
     "batch_norm", "maximum", "minimum", "mha", "cross_attention", "geglu", "sdpa",
     "pixel_shuffle", "pixel_unshuffle", "topk", "sort", "cross_product",
--- a/aneforge/_compile.py
+++ b/aneforge/_compile.py
@@ -76,6 +76,8 @@ class _Emitter:
     self.blob = BlobWriter()
     self.lines: list[str] = []
     self._split_cache: dict = {}       # (src, num_splits, axis) -> emitted part names, so split is emitted once
+    self._n_weights = 0
+    self._n_rejected: dict[str, int] = {}   # encoding -> count of weights that fell through
 
   def line(self, text: str) -> None:
     self.lines.append("        " + text)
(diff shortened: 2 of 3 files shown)


sbryngelson:
Pushed a small follow-up (3aa0dd8): gate the warning to an explicit single encoding and skip `compress="auto"`.

`auto` is documented to fall through by design (sparse → int4 → int8 → fp16), so a fallthrough there isn't a rejection to warn about. It also runs several streams, so one weight could land in two reject buckets and push the reported count above the weight count. Restricting to a single explicit encoding fixes both, and keeps the count exact. Added a `compress="auto"` regression test. Full suite green (29 passed).

Thanks for the PR!

PR #262, opened 2026-09-01
linalg: add kron

### LINKED ISSUE #126: linalg: kron (Kronecker product)
`kron(A, B)` for 2-D `A[m, n]` and `B[p, q]` produces `[m*p, n*q]` and is an outer product plus a reshape -- all layout/broadcast primitives already present, no gather. Build it as `A[:, None, :, None] * B[None, :, None, :]` (a rank-4 broadcast-multiply) then `reshape` to `[m*p, n*q]`, composing `reshape` and elementwise `mul` the way `aneforge/einsum.py` does its outer products.

Add to `aneforge/linalg.py`; oracle is `np.kron`. Each output element is a single product (no summation), so the extraction is bit-exact in fp16 -- assert `array_equal`, not a tolerance, like the pure-diagonal einsum test does. Cover a couple of non-square shapes.

axiom-of-choice:
Adds `kron(A, B)` to `aneforge.linalg`, built as a broadcast-multiply of expanded views plus a reshape (no new kernel):

- `A.expand_dims((1, 3)) * B.expand_dims((0, 2))` → `[m, p, n, q]`, then reshape to `[m*p, n*q]`.

Closes #126.

**Testing**
- `test_kron` in `tests/test_linalg.py` (two non-square shapes) compares against `np.kron`; on-device M2 Pro: passed (max abs err 4.9e-4, ~half ULP fp16).
- Off-device suite: 407 passed, 2 skipped.
- `ruff`, pylint 2-space gate, `pyright aneforge`, `compileall` clean.

### DIFF
--- a/aneforge/linalg.py
+++ b/aneforge/linalg.py
@@ -528,6 +528,7 @@ __all__ = [
   "qr", "cholesky", "lu", "lu_pivoted", "solve", "solve_triangular", "inv", "lstsq", "expm",
   "eigh", "eigvals", "generalized_eigh", "dominant_eig",
   "svd", "dominant_svd", "svdvals_topk", "randomized_svd", "pca",
+  "kron",
 ]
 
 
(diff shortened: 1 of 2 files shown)


sbryngelson (approved):
Verified on device (M5): the kron test passes, and the full linalg suite is unaffected.

The expand_dims/broadcast-multiply/reshape construction is the right shape for this, and
`_check_precision=False` matches the existing convention in this module (linalg.py:36, 360, 716).
Nothing to change.

One scheduling note: #259, #260 and this all append to the same `__all__` block and the same tail
of linalg.py, so whichever lands first will force a rebase on the other two.

PR #264, opened 2026-09-07
dsp: add sawtooth, square, and chirp signal generators

axiom-of-choice:
### Summary

Closes #167. Adds three host-side numpy waveform generators to `aneforge/dsp.py`, matching the existing window-function style near the top of the file:

- `sawtooth(t, width=1.0)` — periodic sawtooth/triangle, period `2*pi`. `width` sets the rising fraction (1 = rising ramp, 0.5 = triangle, 0 = falling). Out-of-range `width` yields NaN, matching scipy.
- `square(t, duty=0.5)` — periodic square wave, `+1` for the first `duty` fraction of the cycle, `-1` after. Out-of-range `duty` yields NaN, matching scipy.
- `chirp(t, f0, t1, f1, method="linear")` — frequency sweep from `f0` at t=0 to `f1` at t=t1; linear method (log/quadratic deferred as optional per the issue). Unsupported `method` raises `ValueError`.

All three compute in float64 and return `float32`, matching the window functions. `_selftest` gains four cases comparing against `scipy.signal`; `__all__` is updated.

### Tests

New `tests/test_dsp_signal_generators.py` (14 tests): parametrized against `scipy.signal.sawtooth`/`square` across widths/duties, NaN on invalid width/duty, chirp vs scipy, and the unsupported-method guard. These are pure numpy generators, so the self-test runs off-device:

- `_selftest` exit 0, relerr: sawtooth 2.0e-16, square 0.0, chirp 2.1e-8
- `pytest tests/test_dsp_signal_generators.py` 14/14 passed
- `pytest -m "not requires_ane" -q` 421 passed, 2 skipped
- `ruff check`, the pylint 2-space indent gate, and `compileall` clean

pyright reports only the pre-existing environment import-resolution errors (numpy/scipy not resolved in the local venv; identical 70 errors on clean `main`).

No on-device run needed: these functions never dispatch to the ANE.

### Risks

- `chirp` implements only `method="linear"` by design (issue marks log/quadratic optional); a later issue can extend it.
- NaN on out-of-range `width`/`duty` is a deliberate parity choice with `scipy.signal`, not a silent clamp.

### Post-merge

None.

PR #261, opened 2026-09-01
fft: add irfft as the inverse real FFT (fixes #176)

axiom-of-choice:
Fixes #176. Adds `aneforge.fft.irfft(X_re, X_im, N)` as the inverse real FFT, the natural completion of the existing `rfft`.

**Approach:** thin wrapper over the existing `ifft_plan` machinery, as the issue suggests. Runs the on-engine inverse FFT over the Hermitian-symmetric spectrum and returns the real part. `ifft` already applies the 1/N normalization, and for a Hermitian spectrum the imaginary part is ~0 (numpy.fft.irfft also returns only the real part), so the real-signal contract holds with no new kernel or plan.

**Changes**
- `aneforge/fft.py`: `irfft` function + export in `__all__`.
- `aneforge/fft.py` `_selftest`: round-trip `irfft(rfft(x)) ~= x` across N=128 and N=512, plus a `numpy.fft.irfft` oracle over the half spectrum.
- `tests/test_fft.py` (new, `requires_ane` like `test_fft2.py`): 8 tests covering `fft`/`ifft`/`rfft`/`irfft` against numpy, the round-trip, the numpy oracle, and that `irfft` returns a real array.

**Checks run**
- On-device (M2 Pro / macOS): `_selftest` PASS, `tests/test_fft.py` 8 passed.
- Off-device: `pytest -m "not requires_ane"` 407 passed, 2 skipped. `ruff check`, the pylint 2-space gate, `pyright aneforge/fft.py tests/test_fft.py` (0 errors), and `compileall` all clean.

CI runs the off-device subset; the transform itself dispatches, so the on-device `_selftest` and `tests/test_fft.py` were run locally and pass on M2 Pro.

PR #265, opened 2026-09-10
dsp: add hilbert (analytic signal)

### LINKED ISSUE #130: dsp: hilbert (analytic signal)
The analytic signal is `ifft(fft(x) * h)`, where `h` is the one-sided step (`h[0]=1`, `h[1..N/2-1]=2`, `h[N/2]=1`, rest `0`) -- built entirely on the `aneforge/fft.py` plans already present. Add `hilbert(x)` to `aneforge/dsp.py` returning the complex analytic signal (real = `x`, imag = the Hilbert transform), assembled host-side from the real/imag outputs the way the other fft-based dsp functions do.

Oracle is `scipy.signal.hilbert`; keep the power-of-two length requirement the fft plans impose (document it, as `welch`/`stft` do). Style per `tests/` for the dsp module.

axiom-of-choice:
## What

Adds `hilbert(x)` to `aneforge/dsp.py`: the analytic signal `z(t) = x(t) + i*H[x](t)` built entirely on the existing on-ANE `fft`/`ifft` plans from `aneforge/fft.py`. Composes the one-sided frequency-domain response `h` (`h[0]=1`, `h[1..N/2-1]=2`, `h[N/2]=1`, rest `0`) then IFFT back, exactly as `scipy.signal.hilbert` does one shot. Closes #130.

## Implementation

- Input `x` is real, interleaved...  (no —) single real array, length must be a power of two (the staged on-ANE FFT's constraint, same as `welch`'s guard:padding to another size would shift the spectrum and break the analytic relationship`.
- `fft(x, zeros, N)` gives the full spectrum; elementwise multiply by the real `h`; `ifft` back→analytic signal.
 Both real and imag parts returned as separate numpy arrays (`z_re == x`, `z_im == H[x]`).

## Why this approach

The Hilbert transform is the canonical FFT filter:the real part reproduces the signal, the imaginary part is the ±90-degree shifted version (the *reason* the analytic signal exists). Enforcing power-of-two length keeps behavior identical to scipy at any power-of-2 length,and avoids the padding-induced spectrum shift the staged FFT would otherwise introduce.

## Verification (on-device, M2 Pro)

| Check | Result |
|---|---|
| `pytest tests/test_dsp_hilbert.py` (on-device) | 5 passed — matches `scipy.signal.hilbert` (incl. real part ≈ x, imag part = −cos for sin input, power-of-2 guard成立 |
| `pytest -m "not requires_ane"` |407 passed, 2 skipped |
| `ruff` / `pylint` / `pyright aneforge/dsp.py` / `compileall` | all clean |

## Risks and mitigations

- **Non-power-of-2 lengths raise** — by design (same as `welch`). Documented in the docstring, and covered by a test.
- Numerics are fp16 (`fft`/`ifft` plans are fp16); the test tolerance (relerr < 5e-2) is consistent with the other FFT-based DSP routines.


sbryngelson (commented):
The FFT construction is right and the numbers are good -- verified on device: ruff clean, the dsp
suites pass, and the three new self-test records land at 9.9e-04, 1.1e-03 and 5.4e-04, in line with
the other FFT-based routines in the table. Two things to fix, and they're connected.

The power-of-two restriction isn't needed, and the reason given for it isn't what the FFT does.
`fft(x, xi, N)` builds a plan for exactly N and never pads -- `_factor` takes any N, splitting it
into up to three dense-DFT stages (a prime N degenerates to one dense N x N DFT). The
pad-to-a-power-of-two behaviour is `_next_fft_size` in `fft_convolve`, where it's an optional
convenience for block sizing, not a constraint the transform imposes. Measured with the guard
removed and nothing else changed:

```
   N      factors   imag relerr
 256    [4, 8, 8]     1.085e-03    pow2
 512    [8, 8, 8]     9.322e-04    pow2
 300   [5, 6, 10]     1.732e-03    NOT pow2
 384    [6, 8, 8]     9.955e-04    NOT pow2
1000 [10, 10, 10]     1.847e-03    NOT pow2
 100    [4, 5, 5]     1.535e-03    NOT pow2
```

But dropping the guard on its own would ship a bug, which is the second thing. The filter always
takes scipy's even-N branch (`h[1:N//2] = 2; h[N//2] = 1`). Odd N has no Nyquist bin, and scipy uses
`h[1:(N+1)//2] = 2` there; applying the even branch to an odd length double-counts the top bin:

```
   N  parity   this filter   scipy's filter
 256    even     9.894e-04        9.894e-04
 300    even     1.696e-03        1.696e-03
 255     ODD     2.568e-02        1.198e-03
 101     ODD     3.024e-02        6.064e-04
 129     ODD     1.098e-02        7.912e-04
```

The power-of-two guard is what's currently hiding this, since a power of two is always even.

Which brings up the tolerance: at `5e-2`, the odd-N result of 2.6e-02 passes. So the test as written
could not have caught the wrong filter even with an odd length in the parametrize list. Actual
accuracy is ~1e-3 at every length I tried, so 5e-3 is the honest bound and still leaves 5x headroom.

The patch below adds the odd-N branch, replaces the power-of-two guard with a `N >= 2` check,
extends the parametrize list to cover even non-powers-of-two and odd lengths, and tightens the three
tolerances. Verified: ruff clean, 31 passed across the dsp tests, self-test still PASS. Reverting
just `dsp.py` and keeping the tests makes 5 of them fail, so they do pin the behaviour.

One thing I left out of the patch because it's your call: `z_re` is the round-tripped real part and
comes back with ~1.1e-03 of error, but `Re(z) = x` holds exactly by construction. Returning `x`
directly would make that output exact for free. The argument against is that the current value
doubles as a round-trip sanity check.

Last, the new test file has no trailing newline -- this is the third one (#261, #264, #265). Ruff
can enforce it: `W292` is clean across the repo today except for `examples/clip_zero_shot.py`, so
adding it to `select` in pyproject.toml costs one trivial fix and then CI catches it instead of a
reviewer. I'll take care of that separately -- no need to do anything here.

```diff
diff --git a/aneforge/dsp.py b/aneforge/dsp.py
index fd2c948..af9cc56 100644
--- a/aneforge/dsp.py
+++ b/aneforge/dsp.py
@@ -329,18 +329,23 @@ def hilbert(x):
   """Analytic signal via the on-ANE FFT: z(t) = x(t) + i * H[x](t).
 
   Returns (z_re, z_im) as numpy arrays: z_re == x and z_im is the Hilbert transform.
-  Length must be a power of two (the staged on-ANE FFT constraints); oracle
-  `scipy.signal.hilbert`.
+  Any length >= 2 works -- the staged FFT plans for exactly N and does not pad (a PRIME N
+  degenerates to one dense N x N DFT, as everywhere else in aneforge.fft).
+  Oracle: `scipy.signal.hilbert`.
   """
   x = np.asarray(x, np.float32).ravel()
   N = x.shape[0]
-  if N & (N - 1):
-    raise ValueError("hilbert: length must be a power of two (the staged on-ANE FFT pads to one, which would shift the spectrum)")
-  # one-sided response in the freq domain: h[0]=1, h[1..N/2-1]=2, h[N/2]=1, rest 0
+  if N < 2: raise ValueError(f"hilbert: need at least 2 samples; got {N}")
+  # one-sided response in the freq domain. Even N has a Nyquist bin to halve, odd N does not --
+  # applying the even branch to an odd length double-counts the top bin (relerr 2.6e-02 vs 1.2e-03
+  # at N=255). Same split as scipy.signal.hilbert.
   h = np.zeros(N, np.float32)
   h[0] = 1.0
-  h[1:N // 2] = 2.0
-  h[N // 2] = 1.0
+  if N % 2 == 0:
+    h[1:N // 2] = 2.0
+    h[N // 2] = 1.0
+  else:
+    h[1:(N + 1) // 2] = 2.0
   # FFT of x, apply the (real) one-sided response, IFFT back to the analytic signal
   Xr, Xi = fft(x, np.zeros(N, np.float32), N)
   y_re, y_im = ifft(Xr * h, Xi * h, N)
diff --git a/tests/test_dsp_hilbert.py b/tests/test_dsp_hilbert.py
index bbebe79..5189432 100644
--- a/tests/test_dsp_hilbert.py
+++ b/tests/test_dsp_hilbert.py
@@ -9,7 +9,8 @@ pytestmark = requires_ane  # hilbert dispatches its FFTs to the ANE
 scipy_signal = pytest.importorskip("scipy.signal")
 
 
-@pytest.mark.parametrize("N", [256, 512, 1024])
+# powers of two, even non-powers of two, and ODD lengths (which need scipy's other filter branch)
+@pytest.mark.parametrize("N", [256, 512, 1024, 300, 384, 255, 101])
 def test_hilbert_matches_scipy(N):
   rng = np.random.default_rng(20260530)
   x = rng.standard_normal(N).astype(np.float32)
@@ -19,10 +20,10 @@ def test_hilbert_matches_scipy(N):
   assert z_im.shape == (N,), f"im shape {z_im.shape} != ({N},)"
   # real part should match the input (analytic signal keeps the signal as its real part)
   re_err = float(np.linalg.norm(z_re - x.astype(np.float64)) / (np.linalg.norm(x.astype(np.float64)) + 1e-30))
-  assert re_err < 5e-2, f"real part relerr {re_err:.3e}"
+  assert re_err < 5e-3, f"real part relerr {re_err:.3e}"
   # imaginary part (Hilbert transform) against scipy
   im_err = float(np.linalg.norm(z_im - ref.imag) / (np.linalg.norm(ref.imag) + 1e-30))
-  assert im_err < 5e-2, f"imag part relerr {im_err:.3e}"
+  assert im_err < 5e-3, f"imag part relerr {im_err:.3e}"
 
 
 def test_hilbert_sine_gives_minus_cosine():
@@ -33,9 +34,9 @@ def test_hilbert_sine_gives_minus_cosine():
   _, z_im = dsp.hilbert(x)
   ref = -np.cos(t.astype(np.float64))
   err = float(np.linalg.norm(z_im - ref) / (np.linalg.norm(ref) + 1e-30))
-  assert err < 5e-2, f"sin->-cos analytic signal relerr {err:.3e}"
+  assert err < 5e-3, f"sin->-cos analytic signal relerr {err:.3e}"
 
 
-def test_hilbert_requires_power_of_two():
+def test_hilbert_rejects_too_short():
   with pytest.raises(ValueError):
-    dsp.hilbert(np.zeros(300, np.float32))
\ No newline at end of file
+    dsp.hilbert(np.zeros(1, np.float32))
```

sbryngelson (approved):
Re-verified on device against current main: ruff clean (including the new W292 rule), 50 passed across the dsp test files, self-test PASS with the three hilbert records at ~1e-3. Good to merge. One note: the PR description still says power-of-two is enforced by design and cites the 5e-2 tolerance -- both changed in the second commit. I'll set the squash message from the code rather than the description so the history reads right.

PR #267, opened 2026-09-12
bench: add M1 MacBook Air roofline results

Rayan-and-beyond:
M1 MacBook Air roofline submission for #137.

- Hardware: MacBookAir10,1, Apple M1 (4+4 CPU, 7-core GPU), 8 GB unified
- macOS 15.6.1 (24G90), AC power, automatic energy mode, thermal nominal
- Mode: `roofline_suite.py --perf` (cliffs + headline perf incl. watts via
  powermetrics), run 2026-09-12
- Added: `bench/results/rooflines/roofline-apple-m1-MacBookAir10_1-bb9e71be3ae6-22b6e4fa.json`
- Regenerated: `bench/results/ROOFLINES.md` (+ `rooflines.json`, same command)
- Validation: `aggregate_rooflines.py --check` passes; table grows 6→7
  machines, 9→10 submissions. New row shows the documented pre-A16
  clamp @ 4094, consistent with the other M1-family submissions.


sbryngelson (approved):
Thanks -- this is a clean submission, and a useful one: another pre-A16 datapoint bracketing the
4094 crop-DMA clamp, from a chassis we didn't have.

What I checked:

- `aggregate_rooflines.py --check` passes, and regenerating from the JSONs reproduces the committed
  `ROOFLINES.md` and `rooflines.json` byte-for-byte, so nothing was hand-edited.
- Schema matches all nine existing submissions. Run from a clean checkout at `caeef8e` (current
  main, `dirty: false`), all four perf scripts returned 0 -- which is more than several existing
  rows can say.
- The numbers sit where an M1 should relative to the M1 Max (same 16-core ANE):

```
                        GEMM TF/s   BW GB/s   decode tok/s
M1 Air        (this)       4.33       5.40        278
M1 Max     (2 runs)     4.56/4.59   6.9/7.4       244
```

  5% under the Max on GEMM is a fanless chassis being a fanless chassis. Cliffs are identical to
  every other pre-A16 machine.

If you compared against the existing M1 row (MacBookPro17,1: 2.70 TF/s, 0.86 GB/s) and wondered,
that one is the outlier, not yours -- it's the first-ever run of the suite from Aug 1, its decode
script failed, and its headline came through the fallback extraction path. Yours is the better M1
datapoint.

PR #269, opened 2026-09-16
onnx: coverage_report preflight for unsupported ops

### LINKED ISSUE #96: onnx: coverage preflight report
Importing a model with unsupported ops fails one op at a time: NotImplementedError on the first missing op, fix or work around it, hit the next. A preflight that walks the whole model first would report everything at once: `coverage_report(model)` in `aneforge/onnx.py` walks `model.graph.node` (and If/Loop subgraph attributes) against the `_ONNX` handler table and returns the unsupported op types with counts, so the error message can say 'this model needs 3 ops the importer does not have: X (4 nodes), Y, Z'.

Pure Python over the existing table -- no compile, no dispatch. Wire it into the `NotImplementedError` path of `onnx_to_tensor` so the first-failure message includes the full list, and expose it publicly for people evaluating models.

Test with hand-built models via the `_model` helper in `tests/test_onnx.py` (mixed supported/unsupported nodes, an unsupported op inside a Loop body). Build-level only, so it runs off-device.

No Apple Silicon needed for this one: the tests are host-side comparisons that run in CI, so it can be done from any machine.

axiom-of-choice:
Closes #96.

`coverage_report(model)` walks `model.graph.node` and any `If`/`Loop` subgraph attributes against the `_ONNX` handler table and returns the unsupported op types with their node counts. It is exported as `af.coverage_report`.

`onnx_to_tensor` now appends that list to the first-failure `NotImplementedError`, so instead of failing one op per attempt the message reports all of them:

```
ONNX op 'Foo' not supported; this model needs 2 ops the importer does not have: Bar, Foo (2 nodes)
```

## Verification

- New `tests/test_onnx_coverage.py` (off-device, build-level): mixed supported/unsupported nodes, an unsupported op inside a `Loop` body, both `If` branches, an all-supported model (empty report), and the aggregated error message.
- `pytest -m "not requires_ane"`: 467 passed, 2 skipped.
- `ruff check`, `pylint` (repo flags), `compileall`: clean. `pyright aneforge/onnx.py aneforge/__init__.py`: 0 errors.

Pure Python over the existing table: no compile and no dispatch, so the whole thing runs in CI off-device.

## Local run

```python
import aneforge as af
from onnx import helper, TensorProto
def _vi(n, s): return helper.make_tensor_value_info(n, TensorProto.FLOAT, s)
nodes = [helper.make_node("Relu", ["x"], ["a"]),
         helper.make_node("Foo", ["a"], ["b"]), helper.make_node("Foo", ["b"], ["c"]),
         helper.make_node("Bar", ["c"], ["y"])]
g = helper.make_graph(nodes, "g", [_vi("x", [1, 4])], [_vi("y", [1, 4])])
m = helper.make_model(g, opset_imports=[helper.make_opsetid("", 13)]); m.ir_version = 9

print(af.coverage_report(m))
try:
    af.onnx_to_tensor(m)
except NotImplementedError as e:
    print(e)
```

```
>>> af.coverage_report(model) = {'Foo': 2, 'Bar': 1}
>>> onnx_to_tensor now raises:
    ONNX op 'Foo' not supported; this model needs 2 ops the importer does not have: Bar, Foo (2 nodes)
```


sbryngelson (approved):
Verified on device: ruff clean, 193 passed across test_onnx_coverage.py and test_onnx.py.

The part worth calling out is that the `except NotImplementedError` wrapper doesn't only fire for
missing ops -- handlers raise it for supported ops in unsupported configurations too (3-D Hardmax,
from #263). Guarding on `if missing:` gets all three cases right, which I checked directly:

```
supported op, bad config   ONNX Hardmax: only 2D [C, W] inputs supported on the ANE
genuinely missing op       ONNX op 'Foo' not supported; this model needs 1 op ...: Foo
both at once               ONNX Hardmax: only 2D [C, W] ...; this model needs 1 op ...: Foo
```

The third is better than I expected -- you get the immediate blocker and the one waiting behind it
in a single message, which is the whole point of a preflight. Nothing to change.

PR #270, opened 2026-09-16
dsp: run the self-test battery under pytest and cover the public guards

### LINKED ISSUE #127: tests: run the module self-test batteries under pytest

axiom-of-choice:
Closes #127 (dsp; the einsum slice landed in #200).

`dsp._selftest()` covers `fir_filter`, `fft_convolve`, `freq_filter`, `spectrogram`, `autocorrelate` and `iir_filter`, none of which any test in `tests/` touched. Nothing ran it: `python -m aneforge.dsp` was manual and `pytest` did not know about it.

- `tests/test_dsp_selftest.py`: runs `dsp._selftest()` and asserts the gate. Gated on `requires_ane` because the battery dispatches FIR/FFT ops to the engine. Passed locally on M2 Pro (worst GOOD relerr 1.27e-03).
- `tests/test_dsp_guards.py`: pins the public guards that reject before any dispatch, so they run off-device in CI. The `welch` and `hilbert` guards move there from the `requires_ane` modules, where CI was skipping them.

## Not done

`linalg.main()` is a printed report with no pass/fail, already covered numerically by `tests/test_linalg.py` (115 tests), so it is left alone. Say the word if you would rather drop it.

Same class of gap, untouched here: `tests/test_spectral_sci.py` and `tests/test_lapack.py` are named like tests but collect zero (they are script batteries with a `main()`).

## Verification

- `pytest -m "not requires_ane"`: 436 passed, 2 skipped (was 426).
- `pytest tests/test_dsp_guards.py tests/test_dsp_selftest.py`: 11 passed, the battery on the ANE.
- `ruff check`, `pylint` (repo flags), `compileall`: clean. `pyright` on the new files: 0 errors.

## Local run

```
$ pytest -m "not requires_ane" tests/test_dsp_guards.py -q
..........                                                               [100%]
10 passed in 0.23s

$ pytest tests/test_dsp_selftest.py -q        # the battery on the ANE
.                                                                        [100%]
1 passed in 2.72s
```


sbryngelson (approved):
Verified on device: ruff clean, 59 passed across the dsp test files, and `_selftest()` does return 0
(not None), so the new assertion is actually checking something.

This is the most valuable of the current batch, for a reason that isn't obvious from the diff: CI
runs no ANE tests at all, so until now every dsp guard was only exercised on a maintainer's Mac.
Moving them into an ungated module means `get_window`, `fir_filter`, `freq_filter`, `correlate`,
`autocorrelate`, `hilbert` and `welch` now have their input validation covered on every PR. That is
a real widening of what CI can catch, not just a test reshuffle.

One wording nit, take it or leave it: the `test_dsp_selftest.py` docstring says "iOS-like CI cannot
reach the ANE" -- "hosted CI" is what's meant; there's no iOS in the picture.

PR #271, opened 2026-09-16
bench: add Apple M4 Max rooflines

n0madic:
## Summary
- add a full Apple M4 Max (Mac16,6) roofline submission
- regenerate the Markdown and machine-readable roofline summaries

## Validation
- On-device: `PYTHONPATH=. python3 bench/roofline_suite.py --contributor n0madic --perf-full` passed on Apple M4 Max (Mac16,6), macOS 26.7 (25G229); all six benchmark stages returned 0.
- `python3 bench/aggregate_rooflines.py --check`
- `git diff --check`

The submitted run was on AC power, in Automatic energy mode.


sbryngelson (approved):
Thanks -- merged-ready, and the first M4 Max in the table.

Checked: `aggregate_rooflines.py --check` passes, regenerating from the JSONs reproduces the
committed `ROOFLINES.md` and `rooflines.json` byte-for-byte, the schema matches all ten existing
submissions, and all six perf scripts returned 0.

The numbers land where they should. Same 16-core ANE generation as the base M4, but with the Max
memory system behind it:

```
             GEMM TF/s   BW GB/s   perf/W
M4              5.44       7.67     1112
M4 Max (this)   8.90      23.0      1103
M5              8.71      20.2      1184
```

Compute tracks the bandwidth rather than the core count, which is the pattern the rest of the table
shows. Cliffs are `exact (no clamp)`, correct for post-A16 silicon and consistent with M4/M5.

Two notes, neither blocking and neither needing a re-run: the run was from a dirty tree (several
existing rows are too), and `aneforge_version` came through as `0+unknown`, which usually means the
checkout had no git tags -- the commit itself was recorded fine. Decode is blank because it ran
under `--perf-full`; the aggregator already handles that and added you to the footnote automatically.

PR #273, opened 2026-09-16
examples: the Mandelbrot set by escape-time iteration on the ANE

### LINKED ISSUE #174: examples: Mandelbrot / Julia escape-time on the ANE
An eye-catching, self-contained demo: the Mandelbrot (and/or a Julia) set by **escape-time iteration on the engine**, in the family of `examples/fluid_vorticity.py` / `examples/reaction_diffusion.py`.

**Approach:**
- Build a coordinate grid of `c = cr + i*ci` over the complex plane. Complex numbers are carried as two real tensors `zr, zi` (the ANE is real-valued; `aneforge/fft.py` already carries spectra this way).
- One iteration is elementwise: `zr, zi = zr*zr - zi*zi + cr, 2*zr*zi + ci`, then an escape mask where `zr*zr + zi*zi > 4`. Accumulate the escape iteration count with `select` / `maximum` (in `graph.py`). Each iteration is one on-engine forward pass over the whole grid.
- After the fixed iteration budget, map the smooth iteration count through a colormap and save a PNG (PIL, like the sibling demos). Optionally sweep a Julia `c` to make a short animation.

**Deliverable:** `examples/mandelbrot.py` (one file, docstring), plus an `examples/README.md` entry.

**Testing:** spot-check a handful of grid points against a plain numpy escape-time reference. Requires any Apple Silicon Mac. Follow `reaction_diffusion.py` for the render, `_common.py` for setup.

axiom-of-choice:
Closes #174.

The Mandelbrot set by escape-time iteration, in the family of `reaction_diffusion.py`.

- The complex orbit `z -> z*z + c` is carried as two real tensors `(zr, zi)` (the ANE is real-valued, as `aneforge/fft.py` does for spectra).
- One iteration is one program over the whole grid: `zr, zi = zr*zr - zi*zi + cr, 2*zr*zi + ci`, then `select` freezes each point once `|z|^2` leaves the escape radius, so its count stops and `z` stays finite.
- Compiled once, re-dispatched `MAX_ITER` times. The smooth count is recovered host-side from the frozen `z`, whose escapee is `z*z + c`, so `log(log|z|)` is finite everywhere it is used.

Writes `docs/assets/mandelbrot.png` (640x640, ~114 KB).

## Verification

- `tests/test_mandelbrot_example.py` (requires_ane): on chosen points away from the boundary the on-engine bounded-iteration counts match `numpy_escape` exactly; four interior points never escape and four exterior points leave early.
- Full run on M2 Pro: 120 iterations at 2.4 ms/step, 12.5% of the plane bounded, PASS.
- `ruff check` and `pylint` (repo flags) on the new test: clean.

## Note

Right at the set boundary the orbit is chaotic, so a 1-ULP fp16 rounding can move the escape iteration by one (4 of 144 points on a 12x12 probe). The test picks points away from it; the rendered image is unaffected.

## Local run

```
$ python3 examples/mandelbrot.py

  ANEForge   - the Mandelbrot set on the Apple Neural Engine
  z -> z*z + c as two real tensors, one on-engine forward pass per iteration

  compile OK one orbit-step program (elementwise, compiled once, re-dispatched per iteration)
  iterate OK 120 orbit steps on a 640x640 grid (0.4s wall, 3.53 ms/step)
  set     12.5% of the plane stayed bounded (the interior is black)
  energy  ~0.4s of ANE step time at the measured ~1.48 W rail ~ 0.6 J for the whole render

  OK wrote docs/assets/mandelbrot.png
  PASS: the ANE ran 120 escape-time iterations on the engine and the set resolved
```


sbryngelson (approved):
Verified on device: ruff clean, tests pass, and the example runs end-to-end -- 120 iterations,
12.5% of the plane bounded, writes the PNG, exits 0.

The care in the test is the right call and worth keeping. Escape-time near the boundary is chaotic,
so a 1-ULP fp16 difference moves the escape iteration and an exact-equality assertion would be
flaky; picking points well inside and well outside the disk, and saying so in the comment, is how
this should be tested. I ran it three times to confirm it's deterministic, and it is.

The freeze-on-escape design is also what makes `smooth()` work -- keeping the last bounded z means
the continuous escape count has a finite log everywhere it's used, rather than a NaN wherever the
orbit ran away. The numpy reference implements the same freeze independently, so the comparison
isn't circular.

Nothing to change.

PR #274, opened 2026-09-16
examples: the Lorenz attractor integrated with RK4 on the ANE

### LINKED ISSUE #175: examples: Lorenz attractor integrated on the engine
A small scientific showpiece to sit alongside `examples/heat_equation.py`, `examples/poisson_spectral.py`, and `examples/nbody.py`: integrate the **Lorenz attractor on the engine** and plot the butterfly.

**Approach:**
- State is the 3-vector `(x, y, z)`; the Lorenz vector field `dx=sigma*(y-x)`, `dy=x*(rho-z)-y`, `dz=x*y-beta*z` is a handful of elementwise ops.
- One RK4 step is a small graph (four field evaluations + the weighted combine). Roll out N steps, dispatching the step each iteration, collecting the trajectory host-side.
- Render the trajectory (e.g. x-z projection) to a PNG with matplotlib or PIL, matching how the other scientific demos save their figures.

**Deliverable:** `examples/lorenz.py` (one file, module docstring), plus an `examples/README.md` entry.

**Testing:** compare the first ~100 steps against a numpy RK4 reference (chaotic systems diverge later, so check the short-horizon trajectory, not the endpoint). Requires any Apple Silicon Mac. `_common.py` has the env/scaffolding; `nbody.py` is a good structural template for an on-engine integrator.

axiom-of-choice:
Closes #175.

The Lorenz attractor with one RK4 step as ONE fused program, in the family of `nbody.py` / `heat_equation.py`.

- The state is a `[B,3]` batch of trajectories; the RK4 step splits it into x/y/z columns with `split`, evaluates the field four times and recombines with `concat`, so nothing round-trips to the host.
- Compiled once and re-dispatched over the rollout, matching the other on-engine integrators.
- Renders `docs/assets/lorenz.png` (x-z projection of 8 nearby initial conditions).

## Local run

```
$ python3 examples/lorenz.py

  ANEForge   - the Lorenz attractor on the Apple Neural Engine
  RK4 (four field evaluations) fused into one program, 8 trajectories per dispatch

  compile OK one RK4-step program (65 ANE ops, compiled once)
  integrate OK 5000 RK4 steps x 8 trajectories (t = 50, 0.22 ms/step)
  check    first 100 steps vs fp32 numpy RK4: max relerr 4.85e-03
  extent   x spans 38.1 (the two lobes fired), ~1.08s of ANE step time ~ 1.60 J at the 1.48 W rail

  OK wrote docs/assets/lorenz.png
  PASS: the ANE integrated 8 Lorenz trajectories (5000 RK4 steps) and the butterfly resolved
```

## Verification

- `tests/test_lorenz_example.py` (requires_ane): the first 100 steps of all 8 trajectories match a fp32 numpy RK4 reference (max abs diff < 0.5 on states of magnitude ~30). Chaos diverges later by design, so the check is short-horizon.
- `ruff check` and `pylint` (repo flags) on the new test: clean.

The `DispatchFloorWarning` at compile is expected for a single small RK4 step (`nbody.py` emits the same); each dispatch is tiny and dominated by the fixed ANE call cost.


sbryngelson (approved):
Verified on device: ruff clean, tests pass, example runs end-to-end -- 8 trajectories, 5000 RK4
steps, x spans 37.8 (both lobes), writes the PNG, exits 0.

The vector field and the RK4 combination are both correct, and fusing all four stages into one
program with no host round-trip is a genuinely good demonstration of the point.

I wanted to know whether `HORIZON = 100` with a 0.5 bound was well-chosen or just happened to pass,
since a chaotic system will eventually break any such test. Measured error growth:

```
step     25     50     75    100    150    200    300    400
err   0.095  0.103  0.097  0.098  0.091  0.154  0.141  0.259
```

0.098 against a 0.5 threshold is 5x headroom, and it stays flat well past the horizon rather than
sitting just under the line -- so the test is calibrated, not lucky. Three repeat runs were
identical. Comparing fp16 against fp32 from the same fp16-rounded initial state is also the right
setup; it isolates integration error from initial-condition divergence.

Nothing to change.

